PR #3, opened 2017-10-22
updated authors

julskast:


Sydriax:
Note: Season is misspelled.

julskast:
Thanks for the catch @Sydriax !!

PR #88, opened 2017-10-26
Followed go style guide recommendations to clear lint warnings

jerold:
Minor house cleaning in the go starter kit. This clears all warnings that might otherwise clutter a go bot creator's analyzer window.


Snaar:
What do all of those "// Blah ..." do out of curiosity? Is that something Go linter forces you to do?

Snaar:
Similarly, does Go have canonical convention on Id vs ID? Some languages do. If Go doesn't, I'd prefer to keep it consistent with other kits with "Id".

jerold:
Go wants you to have a comment on any public function or value (anything starting with a Capital letter), thus the empty comments. It warns if Id isn't ID also.

hypirion:
I think it's nice to have it linted, but adding `// Blah ...` doesn't _really_ do what golint wants you to do though: The public functions and types are still not documented.

Perhaps it would be an idea to properly document them, or remove the `// Blah ...` comments to get someone to document them properly later on?

jerold:
@hypirion totally.

I'm not sure if there is a best practice for commenting language-specific implementations of a common library. Is it tedious to keep comments consistent between all language implementations, or should the go comments direct a user to other Halite documentation?

I get a shirt out of this PR for [Hacktoberfest](https://hacktoberfest.digitalocean.com/) so I'm happy to provide some more legit comments when I get home tonight.

Snaar:
If you create separate pull request just for the Id/ID I'll merge it in right away, it sounds golang does have strong opinion about it. That will bring you one pull request closer to your goal :P

jerold:
Pulled the commenting into https://github.com/HaliteChallenge/Halite-II/pull/88... although, I found some time while compiling other stuff to knock out some comments, and could have probably just done that here... :P

harikmenon:
@jerold is this still required? Or do you want to fix the merge issues before i commit? BTW thanks for the contribution!

jerold:
@harikmenon Let me resolve the merge, one sec :)

jerold:
@harikmenon ready for re-review. :)

PR #89, opened 2017-10-26
py-client: make iteration order over entities in map deterministic

Snaar:
Further progress on the quest for no quant diffs between client languages.

### DIFF
--- a/airesources/ML-StarterBot-Python/hlt/entity.py
+++ b/airesources/ML-StarterBot-Python/hlt/entity.py
@@ -1,8 +1,8 @@
-import math
-
-from . import constants
 import abc
+import math
 from enum import Enum
+from collections import OrderedDict
+from . import constants
 
 
 class Entity:
@@ -104,7 +104,7 @@ class Planet(Entity):
         self.health = hp
         self.owner = owner if bool(int(owned)) else None
         self._docked_ship_ids = docked_ships
-        self._docked_ships = {}
+        self._docked_ships = OrderedDict()
 
     def get_docked_ship(self, ship_id):
         """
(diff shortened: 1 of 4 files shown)


Snaar:
It's a bit nuanced - OrderedDict would make iteration identical to order entities were mentioned on the update string from the engine, not necessarily same as if you sort it. I'm not sure if this has that much overhead. In Java using LinkedHashMap vs HashMap isn't such a big deal - they are both horribly slow.

Snaar:
In cpp bot I went for vector that has all the data and then there is separate map if you really do want to look stuff up using IDs.

Snaar:
Hm. Internet suggests that Python 3.6 dict is already sorted by insertion order out of the box. I've been testing on 3.5 locally. What version do we use on the workers?

j-clap:
I haven't profiled it, but I don't assume it'd be much slower, especially considering ships are <<1000, so I'm assuming worst case a 10x iterations diff, but those would be less heavy than the heavier part (create/delete). Anyway, not terrible, shouldn't affect this runtime much.

j-clap:
3.5 lat I checked, but we were about to update

Snaar:
Ok so this possibly becomes moot if update to 3.6 happens. Let me get 3.6 and test first and then there would be no need for this change.

Janzert:
Default python version used to run bots on workers is now 3.6. So dictionaries should be preserving insertion order.

Python 3.5 is still used for running the actual worker scripts though, but that makes no difference for this.

j-clap:
Thank you @Janzert. As such I think we can close this one

Snaar:
Yeah, sweet, just tested, works as advertized.

PR #100, opened 2017-10-26
A few fixes to the starter go bot

bovard:
1. corrected ClosestPointTo to actually return the clostest point to the target object
2. bounded thrust to make sure the degrees were between 0 and 360
3. Fixed Dock to use the correct constant value
4. Fixed NavigateBasic to stop running into things

### DIFF
--- a/airesources/Go/src/hlt/entity.go
+++ b/airesources/Go/src/hlt/entity.go
@@ -182,7 +182,7 @@ func IntToDockingStatus(i int) DockingStatus {
 
 // Thrust generates a string describing the ship's intension to move during the current turn
 func (ship Ship) Thrust(magnitude float64, angle float64) string {
-	if (angle > 0) {
+	if (angle > 0.0) {
 		boundedAngle := int(math.Floor(angle + .5))
 	} else {
 		boundedAngle := int(math.Ceil(angle - .5))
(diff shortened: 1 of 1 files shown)


airesources/Go/src/hlt/entity.go
```diff
@@ -182,7 +182,13 @@ func IntToDockingStatus(i int) DockingStatus {
 
 // Thrust generates a string describing the ship's intension to move during the current turn
 func (ship Ship) Thrust(magnitude float64, angle float64) string {
-	return fmt.Sprintf("t %s %s %s", strconv.Itoa(ship.ID), strconv.Itoa(int(magnitude)), strconv.Itoa(int(angle)))
+	boundedAngle := int(angle)
```
Snaar: Can you round to nearest instead of round down?

airesources/Go/src/hlt/entity.go
```diff
@@ -182,7 +182,13 @@ func IntToDockingStatus(i int) DockingStatus {
...
+	if boundedAngle < 0 {
```
Snaar: Generally you want to avoid branches (if statements). I don't know if % in golang can result in negative numbers or not - various languages implement it differently. Just in case it can be negative, do boundedAngle = ((boundedAngle % 360) + 360) % 360. Otherwise you can just go with boundedAngle % 360 directly.

Snaar: (% is also more robust than current way that handles only limited input range)

bovard: % in go can't handle negatives

Snaar:
👍

PR #101, opened 2017-10-26
Made hlt_client into a pip package

bovard:
All you need to do now is register it on pip and/or push a new version.

You can install `pip install hlt_client`

After that, instead of doing `./client.py` you can just do `hlt`

How to test:

```
cd tools/hlt_client
pip install --upgrade .
hlt auth
```


bovard:
Someone on your end would need to publish this to pip

harikmenon:
We can do that. Adding @j-clap  to review this first.

PR #110, opened 2017-10-27
Fix damage calculation with multiple targets

lidavidm:
When multiple targets were involved, the engine was incorrectly setting the cooldown on the weapon after applying damage to the first target, causing other ships to not be attacked. The fix here takes advantage of the 2-phase calculation of damage, and uses the fact that we first compute targets and damage before applying it. Now, we apply damage based on the attack event information that we were already tracking, so we know the attack must have happened - we do not have to recheck the cooldown or other matters.

https://forums.halite.io/t/fighting-issue-spreading-damage/261/6

First spotted by @MoreGames on the Halite II forums.

### DIFF
--- a/environment/core/Halite.cpp
+++ b/environment/core/Halite.cpp
@@ -738,10 +738,10 @@ auto Halite::process_events() -> void {
         auto update_damage = [&](hlt::EntityId src, hlt::EntityId target) -> void {
             auto& attacker = game_map.get_ship(src);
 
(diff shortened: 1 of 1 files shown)


harikmenon:
I would like to see 2 reviews of game changes going forward. Not that I dont absolutely trust @j-clap  but lets do two sigma style :)

@Snaar can you take look?

PR #119, opened 2017-10-27
Fix in-browser decompression of very large replays

lidavidm:
This is not the best fix. I'm not sure of the underlying cause. Something more robust might be to use the streaming decompression mode.


libhaliteviz/src/parseWorker.js
```diff
@@ -48,7 +48,7 @@ addEventListener("message", (e) => {
             heapBufferView.byteOffset, bufferBytes);
         if (decompressedSize > 0) {
             // Allocate the buffer to hold the decompressed data
-            const inflatedBuffer = libzstdInstance._malloc(decompressedSize);
+            const inflatedBuffer = libzstdInstance._malloc(decompressedSize + 2048);
```
Snaar: Ahh, the classic "just malloc a bit more" :-)

Snaar:
This behaves strictly better than what is out there right now, right? So might as well.

Issue #106, opened 2017-10-27
Bot compilation issue


harikmenon:
From user

For our reference, here is the trace of the error: Traceback (most recent call last): File "worker.py", line 100, in executeCompileTask archive.unpack(bot_path) File "/home/worker/Halite/apiserver/worker/archive.py", line 26, in unpack shutil.move(os.path.join(tempPath, filename), os.path.join(folderPath, filename)) File "/usr/lib/python3.5/shutil.py", line 542, in move raise Error("Destination path '%s' already exists" % real_dst) shutil.Error: Destination path '/home/worker/Halite/apiserver/worker/tmpqyohabca/bot/bot' already exists

@lidavidm - do we know what is going on?

kmisiunas:
User experiencing this problem here. The submission contains the following files:

https://gist.github.com/kmisiunas/dc47663fe93bd7368561313c91a5d8d6

lidavidm:
Ah. The backend unzips the archive contents into a subdirectory named `bot/`, then moves all the files from the subdirectory to the parent directory. Because the bot upload contains a `bot/` folder, it's complaining that that folder already exists.

lidavidm:
A quick fix would be to rename that folder. On our end, I am not sure why we have this extra step - might be worth getting rid of it to just save the confusion.

kmisiunas:
@lidavidm, thank you for looking into this. I renamed the folder into `mybot` and the problem disappeared, as you suggested. Thanks!

harikmenon:
@lidavidm - do you want make a change to fix this?

lidavidm:
@harikmenon Yeah, would make things less confusing.

Snaar:
If you are touching that part of system, want to take care of #62 as well? :P

lidavidm:
Sure! I swear there was already code for that there, but it seems not.

PR #121, opened 2017-10-27
Nicer compilation

### LINKED ISSUE #62: Support bot submissions with single directory in zip file root
Seems like a common thing is people submitting zip files with single directory inside where all the bot code is. Maybe it's worth supporting that instead of asking them to repackage it?

lidavidm:
Fixes #62, #106

### DIFF
--- a/apiserver/worker/worker.py
+++ b/apiserver/worker/worker.py
@@ -104,17 +104,16 @@ def executeCompileTask(user_id, bot_id, backend):
                 name for name in os.listdir(temp_dir)
                 if os.path.isfile(os.path.join(temp_dir, name))
             ]) == 0 and len(glob.glob(os.path.join(temp_dir, "*"))) == 1:
-                singleFolder = glob.glob(os.path.join(temp_dir, "*"))[0]
-                bufferFolder = os.path.join(temp_dir, backend.SECRET_FOLDER)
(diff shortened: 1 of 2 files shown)


apiserver/worker/worker.py
```diff
@@ -104,17 +104,16 @@ def executeCompileTask(user_id, bot_id, backend):
...
-                bufferFolder = os.path.join(temp_dir, backend.SECRET_FOLDER)
-                os.mkdir(bufferFolder)
+                with tempfile.TemporaryDirectory(dir=TEMP_DIR) as bufferFolder:
+                    singleFolder = glob.glob(os.path.join(temp_dir, "*"))[0]
 
-                for filename in os.listdir(singleFolder):
-                    shutil.move(os.path.join(singleFolder, filename), os.path.join(bufferFolder, filename))
-                os.rmdir(singleFolder)
+                    for filename in os.listdir(singleFolder):
+                        shutil.move(os.path.join(singleFolder, filename), os.path.join(bufferFolder, filename))
```
Janzert: It's ok the way it is but the second argument to shutil.move can simply be the destination directory and the source (first argument) will be moved into it. There is no need for the os.path.join with the filename. The same for line 115 below.

Issue #117, opened 2017-10-27
Win condition #2 (a player occupies all planets) not enforced


LEARAX:
Win condition 2, "A player occupies all planets," is not actually enforced.
![screen shot 2017-10-27 at 12 52 44](https://user-images.githubusercontent.com/27088630/32120475-f46f4614-bb15-11e7-804c-ecc0602e3f19.png)
Orange wins in the above screenshot, as shown in [this replay](https://halite.io/play/?game_id=1954063&replay_class=0&replay_name=replay-20171027-183239%2B0000--2161024039-312-208-1509129114).

Expected behavior: Purple would win, as it controls all planets for a frame.

fohristiwhirl:
I believe this is because a planet is only counted as "occupied" for these purposes when one ship has fully docked. In-progress docking doesn't count.

I think current behaviour is OK?

LEARAX:
It’s not communicated in the docs, and the replay viewer shows the planet as being his before the docking finishes.

lidavidm:
@fohristiwhirl is correct, the ship must be fully docked (the visualizer is misleading in this regard). We will update the documentation to make it clear. I believe purple would have won with this condition, but it appears the bot errored out before it finished docking.

LEARAX:
Thanks.

lidavidm:
@LEARAX Thanks for pointing out the ambiguity. I've added a short clarification as part of #122.

PR #122, opened 2017-10-27
Update planet size in spec

lidavidm:
Fixes documentation discrepancy pointed out by @yngvix at https://forums.halite.io/t/planet-radius-range/308. Also clarifies the documentation for the settlement victory condition as pointed out by @LEARAX in #117.


Janzert:
> and the maximum radius is always at least 5 units.

Should that even be mentioned given that with map sizes being used the actual smallest maximum radius is 12.6?

It took me a few reads through to understand what that clause actually meant, just leaving it out seems to make the section easier to grasp.

PR #125, opened 2017-10-28
add match manager adapted from Halite I

smiley1983:
Hi, following the suggestion from @harikmenon in the Halite II forum, I've created a pull request with my match manager and a readme file explaining its basic operation.

Please let me know if there's anything else I should do.

Best regards,


harikmenon:
Thanks for your contribution, this is amazing. Cant wait to play with it

PR #126, opened 2017-10-28
Initial addition of kotlin starter kit

tmseiler:
This is an adaptation of the Java starter kit to Kotlin.

I have included a gradle file to help with first-time setup of the project. Using IDEA, one should be able to just import the gradle file or open the directory as the project.

### DIFF
--- a/airesources/Kotlin/src/kotlin/halite/Collision.kt
+++ b/airesources/Kotlin/src/kotlin/halite/Collision.kt
@@ -1,0 +1,55 @@
+package halite
+
+object Collision {
+    /**
+     * Test whether a given line segment intersects a circular area.
+     *
+     * @param start  The start of the segment.
+     * @param end    The end of the segment.
+     * @param circle: The circle to test against.
+     * @param fudge  An additional safety zone to leave when looking for collisions. (Probably set it to ship radius 0.5)
+     * @return true if the segment intersects, false otherwise
+     */
+    fun segmentCircleIntersect(start: Position, end: Position, circle: Entity, fudge: Double): Boolean {
+        // Parameterize the segment as start + t * (end - start),
+        // and substitute into the equation of a circle
+        // Solve for t
+        val circleRadius = circle.radius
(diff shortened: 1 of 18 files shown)


harikmenon:
@tmseiler , thanks for your contribution. Just 2 comments, can you add a LANGUAGE file at the root and put Kotlin in it so user bots get associated with Kotlin?

Also if you could a bit more documentation to atleast Mybot, that would be awesome

airesources/Kotlin/src/kotlin/halite/Collision.kt
```diff
@@ -0,0 +1,55 @@
+package halite
+
+object Collision {
+    /**
+     * Test whether a given line segment intersects a circular area.
+     *
+     * @param start  The start of the segment.
+     * @param end    The end of the segment.
+     * @param circle: The circle to test against.
```
Snaar: I think you are a bit out of date from latest Java update history - the ':' was cleaned up recently. Can you go through https://github.com/HaliteChallenge/Halite-II/commits/master/airesources/Java to make sure you are caught up with recent changes?

After that I'll keep Kotlin in mind and update it as well if/when touching Java later.

tmseiler: Sure, I can clean this up. But, going forward, wow in sync do we actually want to keep the two starter kits?

Snaar: If it's obviously same code/text then I'll keep it in sync. I.e. if there is like a math bug in Java I'll check all of the kits derived from it and fix it there if bug is still there. But naturally if design of the kit starts to diverge because of the language - that's actually great.

Snaar:
@harikmenon I've talked to @tmseiler on chat - I'll do the cosmetics myself, there are just couple. And then either of us will poke at the adding stuff to gradle build so that LANGUAGE file ends up in zip, but it would be easier for me if code is already in the github.

PR #129, opened 2017-10-28
Cap speed after acceleration, not before

### LINKED ISSUE #114: Non-zero velocities causing docking to fail
Drag isn't guaranteed to reduce velocity to exactly zero. Sometimes a ship's velocity ends up being something like 1e-16 or so.

But the [can_dock() method](https://github.com/HaliteChallenge/Halite-II/blob/2ff792f77a3f485cfaa1253ad72488dac7aef30c/environment/core/Entity.hpp#L184) checks if a ship's velocity is exactly 0.

Therefore, docking can fail for no apparent reason.

lidavidm:
Fixes #114, thanks to @Janzert

### DIFF
--- a/environment/core/Entity.cpp
+++ b/environment/core/Entity.cpp
@@ -36,17 +36,15 @@ namespace hlt {
 
     auto Velocity::accelerate_by(double magnitude,
                                  double angle) -> void {
-        double new_vel_x = vel_x + magnitude * std::cos(angle);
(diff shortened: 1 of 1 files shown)


lidavidm:
Something even safer would be just to bump up the drag constant, so that this branch always applies:

https://github.com/HaliteChallenge/Halite-II/blob/master/environment/core/Halite.cpp#L361

PR #132, opened 2017-10-28
Create a Clojure starter kit

hypirion:
This is a starter kit for Clojure. It was originally derived from the Java starter kit, but has been changed to be more idiomatic Clojure: It uses dynamic vars for global state, properties on ships and planets are accessed via keywords instead of through getters and setters, and functions operating on ships/planets are not defined through interfaces/classes, but instead as standalone functions.

The code should be reasonable documented and should implement every API function mentioned in https://halite.io/learn-programming-challenge/downloads-and-starter-kits/create-new-starter-kit, either directly or in a more Clojure-idiomatic fashion (e.g. `player.all_ships()` is `(get *owner-ships* player-id)` instead.)

This indirectly depends on #131. Without it, you have to create an empty `MyBot.clj` file in the root directory to get uploads to work.


airesources/Clojure/src/hlt/utils.clj
```diff
@@ -0,0 +1,18 @@
...
+  (.write *logger* (str/join " " args))
+  (.write *logger* "\n")
+  (.flush *logger*))
+
+(defn rad->deg
+  "Translates radians into degrees in the interval [0, 360)."
+  [rad]
+  (let [deg-unclipped (Math/round (Math/toDegrees rad))]
+    ;; Clip to ensure value is in [0, 360), as required by the game engine.
+    (int (mod deg-unclipped 360))))
```
Snaar: Oh sweet, clojure actually has mod right.

airesources/Clojure/src/hlt/engine.clj
```diff
@@ -0,0 +1,214 @@
...
+    (concat (filter filter-fn (vals *planets*))
+            (filter filter-fn (vals (get *owner-ships* *player-id*))))))
+
+(def ^:private default-navigation-opts
+  {:max-corrections max-navigation-corrections
+   :avoid-obstacles true
+   :angular-step (/ Math/PI 180.0)
+   :max-thrust (quot max-speed 2)})
+
+(defn navigate-to
```
Snaar: Something I wanted to do for other kits, but maybe you might do here if you feel that's appropriate - I'd move out all the basic/default/suboptimal/etc navigation code to a separate file and make it clear that this code is non-essential and is just to get you started. Java already kinda has "Nagivation" class for it.

hypirion: Yeah, that's a good suggestion. I'll see if I can separate the math and the navigation code as well, so that you can reuse distance functions, intersect functions etc. that aren't necessarily suboptimal.

hypirion:
I have now split up the `engine` namespace into `entitites`, `math` and `navigation`, and hinted that navigation should be replaced with better navigation code when you've got your bot up and running. The math namespace contains functions that are still useful and aren't suboptimal by themselves.

airesources/Clojure/src/MyBot.clj
```diff
@@ -0,0 +1,56 @@
...
+         (e/dock-move ship planet)
+         (navigation/navigate-to-dock ship planet))))))
+
+(defn -main
+  [& args]
+  (initialize-game
+   (io/send-done-initialized)
+   (doseq [turn (iterate inc 1)]
+     (with-updated-map
+      (utils/log "=========== Turn" turn "===========")
```
Snaar: Space after "Turn" might be good.

airesources/Clojure/src/MyBot.clj
```diff
@@ -0,0 +1,56 @@
...
+
+(defmacro with-updated-map
+  [& body]
+  `(let [m# (io/read-map)]
+     (binding [*owner-ships* (:owner-ships m#)
+               *ships* (:ships m#)
+               *planets* (:planets m#)]
+       ~@body)))
+
+(def my-bot-name "mybot") ;; Replace this with your bot name here.
```
Snaar: You sure don't want to give it some personalized default name? :P

airesources/Clojure/src/MyBot.clj
Snaar: Or does it automatically add spaces when concatenating?

hypirion: In Clojure, `println` and friends adds automatic spaces betweeen arguments, so I thought it made sense to do this for the log function as well. In hindsight, it was stupid to not just let it call `println`, so I changed it to do just that.

airesources/Clojure/src/MyBot.clj
hypirion: Well, since you insist :)

Snaar:
👍

PR #136, opened 2017-10-29
environment: add --no-compression flag to disable compression on hlt …

pyBlob:
As my computer isn't that fast and compression takes up most of the time of one game (especially when there are many ships), I added a switch to disable it. This way the files are considerably bigger, but simulation is faster.


Snaar:
Yeah it looked good to me before, I just wasn't sure from feature point of view (as in I had no opinion), but code itself looks fine.

PR #141, opened 2017-10-29
Break ties randomly at end of game (fixes #128)

### LINKED ISSUE #128: Final tiebreak isn't implemented.
The documentation states that the final tiebreak for a game is random. But this isn't actually implemented in the client and it ends up being based on player id (i.e. order of players given on cmdline).

On production this order is close enough to random that it probably makes no practical difference, but fixing it is probably a good idea anyway. It does stand out for players during local testing when the same player keeps winning over and over.

The fix seems would seem to be adding calls to std::random_shuffle with new_ranking before the stable_sort calls at lines 1025 and 1043 in Halite.cpp. It should really be using the seeded generator though and currently that appears to only be available in the map generator.

This issue was pointed out by @Shummie on Discord.

lidavidm:


Janzert:
The new_rankings need to be shuffled just before line 1024 as well. Which also means the generator should be at the beginning of the games.

Also although I don't think it's actually exploitable it should be noted that an adversarial player will almost certainly know the first n shuffles and possibly the complete state of the random generator by looking at the map that was generated. The former because this basically restarts from the same generator state used in mapgen and the latter because mersenne twister isn't cryptographically secure. But this probably doesn't matter since anyone that can force the game into the tiebreak such that it benefits them can also probably just win outright.

lidavidm:
Thanks, forgot about that case. I'm also not sure if it's exploitable, particularly since this is the last-resort tiebreaker.

PR #144, opened 2017-10-29
Ported the C# starter kit to .NET Core 2.0

teo-tsirpanis:


airesources/CSharp/Halite2/hlt/Log.cs
```diff
@@ -17,7 +17,7 @@ public static void Initialize(TextWriter f)
             instance = new Log(f);
         }
 
-        public static void Log(string message)
+        public static void LogMessage(string message)
```
Snaar: I would prefer you keep it as "Log", LogMessage is a bit too verbose and we don't have any other types of logging that need to be distinguished.

airesources/CSharp/rungame.bat
```diff
@@ -1,2 +1,4 @@
-dotnet build
-./halite -d "240 160" "dotnet ./bin/Debug/netcoreapp1.1/csharp2.dll" "dotnet ./bin/Debug/netcoreapp1.1/csharp2.dll"
\ No newline at end of file
+@echo off
+
+dotnet build -c Release
+halite -d "240 160" "dotnet ./bin/Release/netcoreapp2.0/Halite2.dll" "dotnet ./bin/Debug/Release/Halite2.dll"
```
Snaar: Does "/Debug/Release/" in path make sense? Last line.

airesources/CSharp/rungame.sh
```diff
@@ -5,16 +5,15 @@ rm -r *.hlt
...
 
 # start the dotnet build
-dotnet build
+dotnet build -c Release
 
 if [ $? -eq 0 ]
 then
 # run bots like it would do in our servers
-./halite -d "240 160" "dotnet ./bin/Debug/netcoreapp1.1/MyBot.dll" "python3 ../Python3/MyBot.py"
+./halite -d "240 160" "dotnet ./bin/Debug/netcoreapp2.0/Halite2.dll" "dotnet ./bin/Debug/netcoreapp2.0/Halite2.dll"
```
Snaar: Release in the build command but Debug in the ./halite command - looks a bit fishy?

airesources/CSharp/Halite2/hlt/Networking.cs
```diff
@@ -53,8 +53,6 @@ public static void SendMoves(IEnumerable<Move> moves)
 
         private static String ReadLine()
```
Snaar: Can you check if this terminates properly when there is no input from halite? This behaves differently in Linux and Windows - in Linux it seems halite can kill the bot successfully but on Windows bot client library has to kill itself when stdin is closed.

Snaar: I suspect part of this try/catch originally was to exit cleanly, but I'm not sure.

airesources/CSharp/Halite2/hlt/Log.cs
ajesusflores: I would leave it as LogMessage; with Log I'm getting this error: *'Log': member names cannot be the same as their enclosing type*; I had to change Log to LogAction in my local code to make it work.

Snaar: Roger. Makes sense.

airesources/CSharp/Halite2/hlt/Networking.cs
Snaar: This is an example of what cpp lib has to do just for windows: https://github.com/HaliteChallenge/Halite-II/blob/99cec0c7763f6a74ff46d6c3a9d89d5c28638d40/airesources/C%2B%2B/hlt/hlt_in.hpp#L127

Snaar:
I pushed out #145 and removed App.config and also renamed Log to LogMessage ahead of time myself since those were breaking compilation, sorry for inconvenience and thanks for the feedback.

airesources/CSharp/rungame.bat
teo-tsirpanis: Fixed.

Snaar:
I'm not really familiar with this build mechanism - how does it know what files to build? I don't see a list of files to compile anywhere anymore.

Janzert:
Just a note, as currently structured this will still be compiled and run with the Mono runtime on the servers. Although not being a C# user I don't know if the Mono version used on the servers will have a problem with these changes.

Currently the compiler script is distinguishing between Mono and .NET Core runtimes by looking for a MyBot.cs file or a MyBot.csproj file in the root of the upload. Having both will cause an error to be returned.

From what I can tell, in order to have a bot compile and run under the .NET runtime the bot needs to satisfy the following conditions:
 * MyBot.csproj in the root directory of the upload
 * without having a MyBot.cs file there
 * produce a MyBot.dll file when 'dotnet build -c Release -o .' command is run
 * the bot will then be run in games with the command 'dotnet MyBot.dll'

teo-tsirpanis:
> * the bot will then be run in games with the command 'dotnet MyBot.dll'
@Janzert, I think it's better to run `dotnet run -c Release --no-build` instead. The generated DLL is in a `bin` subdirectory. I don't think that hardcoding it is a good idea.

Janzert:
I haven't used C# or any .net language so I don't know what the difference is. The above is simply a description of the way things are currently setup.

If you think it should be changed maybe take a look at changing the appropriate compiler and language entries around lines 288 and 260 respectively in https://github.com/HaliteChallenge/Halite-II/blob/master/apiserver/worker/compiler.py

Snaar:
I can't find dotnet sdk package for Yakkety (16.10) that our workers run on.

Microsoft has this unadvertised repo for Yakkeky that doesn't have anything meaningful in it: https://packages.microsoft.com/repos/microsoft-ubuntu-yakkety-prod/pool/main/d/

They do have releases for 16.04 and 17. We are currently getting 1.1 from https://apt-mo.trafficmanager.net/repos/dotnet-release/pool/main/d/ which has 16.04 and 16.10 for dotnet-dev-1.1.0 but only 16.04 and no 16.10 for dotnet-sdk-2.0.0 and doesn't even have more recent versions of it.

It sounds like we can't actually upgrade from 1.1.0 since workers wouldn't be able to run it.

Snaar:
@harikmenon just for completeness, how painful is it to change the ubuntu version that workers are running on?

Snaar:
Or alternatively if someone can find source of dotnet 2.x for 16.10 that should also be ok.

Janzert:
Updating the worker OS would be fairly risky in terms of possibly breaking already uploaded bots.

harikmenon:
@Snaar agreed with @Janzert. Its hard. Do note that I was able to build and publish a dotnet bot (on NET Core) a month before the competition with @lidavidm on our current workers

teo-tsirpanis:
> Or alternatively if someone can find source of dotnet 2.x for 16.10 that should also be ok.

@Snaar, check out [here](https://github.com/dotnet/core/blob/master/release-notes/2.0/2.0-supported-os.md). Ubuntu 16.10 is an unsupported release, so .NET Core 2.0 does not support it. I would suggest returning to Ubuntu 16.04, an LTS release.

> Updating the worker OS would be fairly risky in terms of possibly breaking already uploaded bots.

True.

lidavidm:
@Snaar aren't we running 17.04?

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 17.04
Release:        17.04
Codename:       zesty
```

in which case we have [official packages](https://github.com/dotnet/core/blob/master/release-notes/download-archives/2.0.0-download.md#ubuntu-1704).

Snaar:
@lidavidm o, i was strictly going by yakkety being used in setup_worker_image.sh - this is good then, I'll file PR to upgrade our dotnet version first, it SHOULD be backwards compatible wiht 1.x and not break anything.

lidavidm:
Yeah, it's a little janky since not all the runtimes had explicit 17.04 repositories.

teo-tsirpanis:
I will propose something more drastic. How about we _remove_ support for Mono (or at least, stop accepting new bots that run on it) and support _only_ .NET Core 2,0? Regarding writing bots, there isn't anything that Mono supports and .NET Core doesn't (and personally, I think that Mono is legacy, now that .NET Core is out). There would be only one .NET Core target that would cover C# 😐, Visual Basic 🤢, and F# 🤩. And it would simply run a `dotnet run -c Release` (`restore` is implied).

Snaar:
So the dotnet 2.0.3 install on linux doesn't seem to come with any of the old frameworks, so our current kit which refers to pre-netcore framework 4.6.1 doesn't build even if you install 1.0 and 1.1 dotnet versions too.

Good news is that there is only one person at all that uses .net core build - https://halite.io/user/?user_id=1318 - and they haven't updated their bot in almost a month and it's not playing any more games. Maybe bot is broken anyway already?

So, I propose not caring if we break any existing netcore bots. I'll go ahead with PR to update workers with latest 2.0.3 netcore. Note that mono bots should be unaffected. Can confirmation I'm not missing anything wrt to my analysis of the situation and plan? @lidavidm?

Snaar:
@harikmenon, what your opinion on this too?

harikmenon:
Sounds good to me.

Snaar:
Just as an update, waiting for #322 to be deployed first and making sure it doesn't have issues. Don't want to break ALL of C# at once :P

PR #149, opened 2017-10-29
Fix running of Haskell bot

rossmacarthur:
- Avoids the following error "Most RTS options are disabled. Link with -rtsopts to enable them."
- I want to run my bot!


rossmacarthur:
Okay updated it to be better. Actually add the `-rtsopts` flag to compiling does the trick. It didn't work when I tested because I didn't delete my *.o files first, I forgot ghc checks the last modified time of source files etc. So it didn't relink them.

PR #156, opened 2017-10-30
Run API servers with Gunicorn (fixes #147)

### LINKED ISSUE #147: Use Gunicorn instead of Flask built-in server
The built-in Flask server is single-threaded, with Gunicorn or something similar, we can serve multiple requests at the same time, hopefully speeding up the API and coordinator (up to a limit). I tried swapping two of the coordinators to Gunicorn with workers=5, bringing up our games/hour from 700 to 930.

lidavidm:
The Flask built-in server is single threaded, and not production ready. Gunicorn lets us service multiple requests at once. This helps alleviate a small bottleneck in running games, though it's not the root cause of the problem that happened 10/29.

### DIFF
--- a/apiserver/requirements.txt
+++ b/apiserver/requirements.txt
@@ -17,6 +17,7 @@ google-cloud-core==0.24.1
 google-cloud-storage==1.1.1
 google-resumable-media==0.0.2
 googleapis-common-protos==1.5.2
+gunicorn==19.7.1
 httplib2==0.10.3
 idna==2.5
 itsdangerous==0.24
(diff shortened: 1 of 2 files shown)


lidavidm:
Seems like we're running just fine without it now, though.

j-clap:
Couldn't we instead use the multithreading flag in Flask? I've used it successfully many times.

harikmenon:
@lidavidm - can you comment on @j-clap comment above?

lidavidm:
I replied in the issue thread - we can do that, just need time to test it out.

harikmenon:
@j-clap what do you want to do here?

j-clap:
Let's try the flask built-in first.

Issue #151, opened 2017-10-30
Starter kits don't let bots take advantage of initial 60s map timeout


Snaar:
Issue is that most (all?) kits send bot name to halite right away, but that's what stops the initial 60s timeout, so right now there is no way to take advantage of the initial 60s timeout for preprocessing the map. Kits need to cache the bot name and delay sending it until getting the map for first turn. Fixed in:

- [x] C++
- [ ] CSharp
- [x] Clojure
- [x] Dart
- [x] Elixir
- [ ] Go
- [x] Haskell
- [x] Java
- [x] JavaScript
- [x] Julia
- [ ] Kotlin
- [x] OCaml
- [ ] PHP
- [x] Python3 & ML-StarterBot-Python
- [x] Ruby
- [x] Rust
- [ ] Scala
- [x] Swift

Snaar:
We got Python and cpp kits fixed. I'll be going through rest one by one over next few days.

hypirion:
The Clojure kit is able to take advantage of the initial timeout if you want it to: `(initialize-game ...)` reads the player + map info, but the bot doesn't send back its name before [`(io/send-done-initialized)`](https://github.com/HaliteChallenge/Halite-II/blob/afe9836e2efa8f377c7ed69d2900682e26a15c7f/airesources/Clojure/src/MyBot.clj#L52) is called.

Snaar:
@hypirion do you mind adding example of initial map analysis so it's clear where to do it in the kit? For example for cpp I've done: https://github.com/HaliteChallenge/Halite-II/commit/76cb289920d053c96d1f12997a878a57bc367ca0#diff-eef5a713e6a41b5875a4f93bcbe7093aR10

hypirion:
@Snaar: Yeah, no problem. I'll see if I can come up with an example.

mw0401:
I still don't understand what "fixed" means. In Python how do access the map information before I submit a bot name?

I can't seem to access game that before I send the command to start my first turn.

this works:
game = hlt.Game("Bot")
Q = game.initial_map

this doesn't:
Q = hlt.Game.initial_map
game = hlt.Game("Bot")

Am I coding it wrong or is it not available, or available differently?

harikmenon:
Closing this out since this is the end of the competition

PR #157, opened 2017-10-30
Add Haskell starter bot

rossmacarthur:


airesources/Haskell/Hlt/Entity.hs
```diff
@@ -0,0 +1,113 @@
...
+-- | Returns the x component of the difference between two Entities.
+dX :: Entity -> Entity -> Float
+dX a b = x b - x a
+
+-- | Returns the y component of the difference between two Entities.
+dY :: Entity -> Entity -> Float
+dY a b = y b - y a
+
+-- | Returns the angle between two Entities.
+getAngle :: Entity -> Entity -> Float
```
Snaar: I suggest you make it explicit if angle is radians or degrees throughout the kit. For example this might be getAngleDeg or getAngleRad.

airesources/Haskell/Hlt/Entity.hs
```diff
@@ -0,0 +1,113 @@
...
+canDock :: Entity -> Entity -> Bool
+canDock s p = getDistance s p <= radius p + dockRadius
+
+-- | Checks whether the line segment between a Ship and an Entity collide with a circle on the GameMap.
+isSegmentCircleCollision :: Entity -> Entity -> Entity -> Bool
+isSegmentCircleCollision a b c = do
+   let dx = dX a b
+       dy = dY a b
+       j = dx**2 + dy**2
+   if j == 0 then (getDistance a c <= (radius c) + 0.75)
```
Snaar: Where is 0.75 coming from?

airesources/Haskell/Hlt/GameMap.hs
```diff
@@ -0,0 +1,36 @@
+-- | A data type representing the game map at a particular frame, and functions that operate on it.
+module Hlt.GameMap where
+
+import qualified Data.Maybe as Maybe
+import qualified Data.Map as Map
+import Hlt.Entity
+
+-- | A GameMap contains the current frame of a Halite game.
+data GameMap = GameMap { myId :: Int
+                       , width :: Float
```
Snaar: Similar comment to max speed - any reason not to use int?

airesources/Haskell/Hlt/GameMap.hs
```diff
@@ -0,0 +1,36 @@
...
+getMyShips :: GameMap -> [Entity]
+getMyShips g = Map.elems $ ships $ Maybe.fromJust $ Map.lookup (myId g) (allPlayers g)
+
+-- | Return a List of all Entities.
+getAllEntities :: GameMap -> [Entity]
+getAllEntities g = getAllPlanets g ++ getAllShips g
+
+-- | Check if there are Entities between two Entities.
+entitiesBetween :: GameMap -> Entity -> Entity -> Bool
+entitiesBetween g a b = any (\e -> isSegmentCircleCollision a b e) $ filter (\e -> e /= a && e /= b) (getAllEntities g)
```
Snaar: Can you add functionality to selectively ignore ships or ignore planets or both? This hasn't propagated to all the kits yet, but Python kit for example has it and I'd like to have it in all the kits eventually.

Snaar: Note that Python defaults to ignoring the ships.

airesources/Haskell/MyBot.hs
```diff
@@ -0,0 +1,44 @@
...
+import Hlt.Navigation
+import Hlt.Networking
+
+-- | Define the name of our bot.
+botName :: String
+botName = "AdamBot"
+
+-- | Log to a file
+info :: String -> IO ()
+info s = appendFile (botName ++ ".log") (s ++ "\n")
```
Snaar: Can you add player id to the file name? Otherwise there will be an issue if two instances of the same bot are playing, right? Other kits do <id>_<botname>.log

Snaar:
Can you address https://github.com/HaliteChallenge/Halite-II/issues/151 as well? If you are not sure what it means I can elaborate more.

airesources/Haskell/Hlt/Entity.hs
```diff
@@ -0,0 +1,113 @@
...
+data Player = Player { playerId :: Int, ships :: Map.Map Int Entity } deriving (Show)
+
+{-|
+  An Entity is an object on the GameMap.
+
+  All Entities have an (x, y) position and a radius.
+
+  Planets and Ships both have corresponding health and owner attributes and various extra attributes.
+-}
+data Entity = Location { x :: Float
```
zxqfl: a data declaration is like a tagged union; it's not the best thing to use here because you want to be able to, say, write a method that returns a `Ship`

but with this setup, you can only write a method that returns `Entity`, which prevents the compiler from doing its job of preventing errors at compile-time

as the `Entity` will need to be converted to a `Ship` at runtime

the better way to express this concept is with typeclasses, which are similar to interfaces in other languages

i have rewritten this file in this way; you will have to change some other files too but i think it will be straightforward to do so

```
class Entity t where
    x :: t -> Float
    y :: t -> Float
    radius :: t -> Float

newtype PlayerID = PlayerID Int
    deriving (Show, Eq)
newtype ShipID = ShipID Int
    deriving (Show, Eq)
newtype PlanetID = PlanetID Int
    deriving (Show, Eq)

data DockingStatus = Undocked | Docking | Docked | Undocking
     deriving (Show, Eq)

data Location = Location { locationX :: Float
                         , locationY :: Float
                         } deriving (Show, Eq)

data Planet = Planet { planetId :: PlanetID
                     , planetX :: Float
                     , planetY :: Float
                     , planetHealth :: Int
                     , planetRadius :: Float
                     , dockingSpots :: Int
                     , production :: Int
                     , planetOwner :: Maybe PlayerID
                     , dockedShips :: [ShipID]
                     } deriving (Show, Eq)

data Ship = Ship { shipId :: ShipID
                  , shipX :: Float
                  , shipY :: Float
                  , shipHealth :: Int
                  , shipOwner :: PlayerID
                  , dockingStatus :: DockingStatus
                  , planet :: Maybe PlanetID
                  , dockingProgress :: Int
                  , weaponCooldown :: Int
                  } deriving (Show, Eq)

instance Entity Location where
    x = locationX
    y = locationY
    radius = pure 0
    
instance Entity Planet where
    x = planetX
    y = planetY
    radius = planetRadius
    
instance Entity Ship where
    x = shipX
    y = shipY
    radius = pure shipRadius

-- | Returns the x component of the difference between two Entities.
dX :: Entity a => Entity b => a -> b -> Float
dX a b = x b - x a

-- | Returns the y component of the difference between two Entities.
dY :: Entity a => Entity b => a -> b -> Float
dY a b = y b - y a

-- | Returns the angle between two Entities.
getAngle :: Entity a => Entity b => a -> b -> Float
getAngle a b = atan2 (dY a b) (dX a b)

-- | Returns the distance between two Entities.
getDistance :: Entity a => Entity b => a -> b -> Float
getDistance a b = sqrt ((dX a b)**2 + (dY a b)**2)

-- | Returns the distance between the edges of two Entities.
getDistanceEdges :: Entity a => Entity b => a -> b -> Float
getDistanceEdges a b = getDistance a b - radius a - radius b

-- | Return the closest point (Circle) to Entity
getClosestPointTo :: Entity a => Entity b => a -> b -> Location
getClosestPointTo s p =
   let a = getAngle p s
       r = radius p + 3
   in Location ((x p) + r * cos(a)) ((y p) + r * sin(a)) 0

-- | Checks if an Planet is owned by a Player.
isOwned :: Planet -> Bool
isOwned p = planetOwner p /= Nothing

-- | Checks if a Ship is undocked.
isUndocked :: Ship -> Bool
isUndocked s = dockingStatus s == Undocked

-- | Checks if a Ship is docking.
isDocking :: Ship -> Bool
isDocking s = dockingStatus s == Docking

-- | Checks if a Ship is docked.
isDocked :: Ship -> Bool
isDocked s = dockingStatus s == Docked

-- | Checks if a Ship is undocking.
isUndocking :: Ship -> Bool
isUndocking s = dockingStatus s == Undocking

-- | Checks if a Planet is fully docked.
isFull :: Planet -> Bool
isFull p = length (dockedShips p) == dockingSpots p

-- | Checks if a Ship is within docking range of Planet.
canDock :: Ship -> Planet -> Bool
canDock s p = getDistance s p <= radius p + dockRadius

-- | Checks whether the line segment between a Ship and an Entity collide with a circle on the GameMap.
isSegmentCircleCollision :: Entity a => Entity b => Entity c => a -> b -> c -> Bool
isSegmentCircleCollision a b c = do
   let dx = dX a b
       dy = dY a b
       j = dx**2 + dy**2
   if j == 0 then (getDistance a c <= (radius c) + 0.75)
   else (do
      let k = -2 * ((x a)**2 - (x a)*(x b) - (x a)*(x c) + (x b)*(x c) + (y a)**2 - (y a)*(y b) - (y a)*(y c) + (y b)*(y c))
          t = min (-k / (2 * j)) 1.0
      ((t >= 0) && ((getDistance (Location ((x a) + dx * t) ((y a) + dy * t) 0) c) <= (radius c) + 0.75)))
```

rossmacarthur: Hi thanks for your input! I'll be sure to update it to use this. One question is it still possible to do something like this? How would this change to fit the new model

```
getAllEntities :: GameMap -> [Entity]
getAllEntities g = getAllPlanets g ++ getAllShips g
```

zxqfl: in general this isn't idiomatic: https://stackoverflow.com/questions/12774056/haskell-list-of-instances-of-a-typeclass

but in this case it seems like the purpose is to get a list of things that you don't want to collide with

simple solution: `getAllEntities :: GameMap -> [Either Planet Ship]`

general solution:
```
data Circle = {x :: Float, y :: Float, radius :: Float}

class Entity t where
    shape :: t -> Circle

getAllEntityShapes :: GameMap -> [Circle]
```

Snaar:
Minor style point: getX isn't very idiomatic. So instead of "getDistance a b" let's do "distance a b".

airesources/Haskell/Hlt/Constants.hs
```diff
@@ -0,0 +1,46 @@
...
+-- | Constant number of turns it takes to dock a Ship.
+dockTurns :: Int
+dockTurns = 5
+
+-- | Constant radius of a Ship.
+shipRadius :: Float
+shipRadius = 0.5
+
+-- | Constant maximum units a Ship can move in one frame.
+maxSpeed :: Float
```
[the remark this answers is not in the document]
rossmacarthur: I made this a float so that it is easy to work with if you want to scale it etc. I floor the value when you make a thrust move

rossmacarthur:
#151 has been addresses too

Snaar:
👍

PR #177, opened 2017-11-01
add OCaml starter package

smiley1983:
Hi, I've made an OCaml starter package, approximately following the Python starter.

The behaviour seems to match that of the Python starter from visual observation, though I haven't tested to make sure it's identical. The navigate function should be exactly the same.

The development repository is here:

https://github.com/smiley1983/halite2-ocaml-starter

### DIFF
--- a/airesources/CSharp/Halite2/hlt/MetadataParser.cs
+++ b/airesources/CSharp/Halite2/hlt/MetadataParser.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Globalization;
 
 namespace Halite2.hlt {
     public class MetadataParser {
@@ -12,8 +13,8 @@ namespace Halite2.hlt {
 
         private static Ship NewShipFromMetadata(int owner, Metadata metadata) {
             int id = int.Parse(metadata.Pop());
-            double xPos = double.Parse(metadata.Pop());
-            double yPos = double.Parse(metadata.Pop());
+            double xPos = double.Parse(metadata.Pop(), CultureInfo.InvariantCulture);
+            double yPos = double.Parse(metadata.Pop(), CultureInfo.InvariantCulture);
             int health = int.Parse(metadata.Pop());
 
             // Ignoring velocity(x,y) which is always (0,0) in current version.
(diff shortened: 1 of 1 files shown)


airesources/OCaml/game.ml
```diff
@@ -0,0 +1,251 @@
...
+  set_planets_docked_ships state;
+;;
+
+let init state name initial_bot_setup =
+  state.my_id <- int_of_string (Networking.get_string ());
+  Debug.setup_logging state.my_id name;
+  Networking.read_width_height state (Networking.get_string ());
+  update_map state;
+  initial_bot_setup state;
+  Networking.send_string name;
```
Snaar: That's not going to work I think for #151 - the 60 second initial timeout ends as soon as you send the bot name over. This has to be done right before first map update.

smiley1983: I seem to be misunderstanding something then... is there anything useful the bot can do with those 60 seconds if it hasn't seen the map state yet?

This code does appear to work with the distributed Halite engine - if I output the state of the planets in that `initial_bot_setup` function, it does have the complete list of planets at that point.

In the Python starter, the `__init__` function in networking.py calls `update_map` with `_send_name = False` before setting `_send_name = True" - to me this seems to have the same result as the OCaml code here, but I'm guessing I've missed something. I'll await clarification.

Snaar: It's possible I misunderstood something. Here is protocol:

- engine sends initial map to bot kit and starts the 60 second timer
- bot kit framework receives initial map and shows it to the bot
- bot author can do whatever it wants with the initial map for 60 seconds, such as prebuild pathfinding-related stuff
- bot author notifies bot kit framework that they are done, either explicitly or asking for first map update
- bot kit framework sends bot name to game engine, thus terminating the 60 second timer
- game engine sends first turn map over - actually identical to initial map - and the normal 2 second turn timer starts

Does this help?

Snaar: Actually, how about you update MyBot to show exactly where the user can take the initial map and spend 60 seconds analyzing it? Sorta like what cpp and Java kits do - take a look at their MyBot examples.

Snaar:
Can you add .mli files? I guess for everything but MyBot and td.ml. The mli files would make it much easier to get an overview of what functions you can use and what types they take and return.

airesources/OCaml/game.ml
smiley1983: That `initial_bot_setup` function in where the user would do their 60 seconds of processing. It's a blank function in MyBot.ml with the comment "Anything you want to do with your opening time".

smiley1983: Regarding .mli files, I'll look into it. I don't normally use them at all, but I'll see if I can do that. It'll take some time.

smiley1983:
Sorry, I messed up the git history of my repo and reset it - closing this pull request before opening a new one. My apologies for that.

PR #185, opened 2017-11-01
Revert pumpkins

### LINKED ISSUE #184: Remove the pumkins :)
It's been fun, but it's time to move on. Could we have the planets back?

lidavidm:
Fixes #184


harikmenon:
@lidavidm , instead of reverting, can we just turn this off?, we can reuse the code for New years and Christmas?

PR #191, opened 2017-11-02
Have might_attack take into account ship radius.

Janzert:
This fixes the problem where attacks weren't happening until ships were within weapon radius, instead of weapon radius + ship radius * 2.

Reported by @fohristiwhirl at https://forums.halite.io/t/what-are-weapon-ranges-supposed-to-be/417

### DIFF
--- a/environment/core/SimulationEvent.cpp
+++ b/environment/core/SimulationEvent.cpp
@@ -172,6 +172,7 @@ auto collision_time(double r, const hlt::Ship& ship1, const hlt::Planet& planet)
 
 auto might_attack(double distance, const hlt::Ship& ship1, const hlt::Ship& ship2) -> bool {
     return distance <= ship1.velocity.magnitude() + ship2.velocity.magnitude()
+        + ship1.radius + ship2.radius
         + hlt::GameConstants::get().WEAPON_RADIUS;
 }
 


fohristiwhirl:
Perhaps we should leave this alone? (Or make the engine consistent but keep effective range of 5?) Changing to 6 would be a big change. I certainly always thought it would be measured centre to centre.

Janzert:
The problem with leaving it alone is that it gives inconsistent results for distances between 5 and 6. Although I haven't actually tested I'm pretty sure in your weapon-test replay, if you had tried to have the ship leave again it probably would get hit. Almost certainly if it had approached to a distance of say 5.5 then left.

Janzert:
Here's a replay showing how the current behavior is inconsistent. Notice that the ship can sit at a distance of 6 without any combat, but when it steps away combat is triggered.


[might_attack_bug.zip](https://github.com/HaliteChallenge/Halite-II/files/1437414/might_attack_bug.zip)

fohristiwhirl:
Then maybe, once the constant is applied, the WEAPON_RADIUS should be set to 4? That might be the most minimal change in terms of gameplay impact?

lidavidm:
The actual attack code was always measuring edge-to-edge, but the broadphase collision (which this fixes) was only measuring center-to-center. But as Janzert points out, this means that you get inconsistent results on the edge of that range when ships are moving. I think the spec states it should be center-to-edge, though (["five units from the center of the ship"](https://halite.io/learn-programming-challenge/basic-game-rules/game-rules-deep-dive#ship-ship-combat--collision)).

fohristiwhirl:
Can we get a version bump if this happens? I know some important bugfixes have occured yet the version is still 1.2 as far as I can tell. This seems even more important, actually affecting combat...

lidavidm:
Yeah, this should bump the version for sure. I'm also thinking we should embed the Git commit hash in the replay, so that we can tell what version generated it.

fohristiwhirl:
"five units from the center of the ship".

I think most people interpreted that as meaning 5 from center to center. (And that was the actual behaviour we saw most of the time, e.g. when we try to close with the enemy.)

So I still think changing the constant to 4 might be good.

lidavidm:
Yeah, the other way would be to leave `might_attack` as-is and change the call to `collision_time` to only give the attack radius, and ignore the ship radii.

fohristiwhirl:
In that case, the constant could stay the same, which is probably less confusing for anyone who ever looks at it. I think that's the minimal change that makes things consistent.

Janzert:
The distance checks are edge to edge for docking and collision, so I'd rather see this become truly edge to edge as well. Looking at 5 recent daewook games between 40-60% of attacks are missed but the median distance for combat is either at or close to 6. So fixing this either way will change the current behavior quite often.

  game 2147864:
    Combat distance minimum 1.04 median 5.97 mean 4.97 maximum 6.00
    Combat missed 857 (50.65%) found 835 (49.35%)
  game 2150980:
    Combat distance minimum 1.06 median 5.40 mean 4.72 maximum 6.00
    Combat missed 653 (61.03%) found 417 (38.97%)
  game 2151474:
    Combat distance minimum 1.02 median 6.00 mean 5.17 maximum 6.00
    Combat missed 2144 (56.41%) found 1657 (43.59%)
  game 2154168:
    Combat distance minimum 1.04 median 6.00 mean 5.12 maximum 6.00
    Combat missed 340 (39.81%) found 514 (60.19%)
  game 2157007:
    Combat distance minimum 1.39 median 6.00 mean 5.05 maximum 6.00
    Combat missed 54 (40.00%) found 81 (60.00%)

Janzert:
One more stat, in the above games combat occurred at a distance of 5 or less in 26-46% of the engagements. So bring the maximum distance down to 5 would either delay or outright avoid quite a bit of the combat that actually took place.

lidavidm:
That's a good point, and thanks for sharing these statistics. This bug mostly affects the case where ships are stationary or near-stationary and are at max range from each other, right? So it sounds like the minimal change is to "increase" the attack range.

fohristiwhirl:
OK I withdraw my objections.

Janzert:
Yeah, the bug causing ships at distances from 5 to 6 with velocity less than 1 to get missed.

Not sure what change you mean by "So it sounds like the minimal change is to "increase" the attack range".

I think the minimal change is to apply this might_attack fix and leave everything else as is. This will add additional combat events that are currently missed, but a smaller number than will change if the weapon range is just calculated center to center.

lidavidm:
Right, that was poorly worded - the change with the least impact is this, which in some cases will "increase" the range at which ships attack, but this is to make it match the other cases.

Janzert:
Yep, that sounds right.

lidavidm:
@j-clap are you able to look at this, or maybe @Snaar could instead?

j-clap:
I'm okay with this change as long as we properly document it to all users. It will change expected behavior so there will be discontent either way.

PR #199, opened 2017-11-02
Fixes the collision time calculation

nomad010:
The collision detection code ignores certain cases. This makes the code over-enthusiastic when saying there is a collision.

Synopsis of fix:
* Calculate t1 and t2 correctly
* The only valid t values are between 0 and 1. Values less than 0 indicate a collision if we went back in time. Values greater than 1 indicate a future collision, but only after the circles would have stopped moving.

### DIFF
--- a/environment/core/SimulationEvent.cpp
+++ b/environment/core/SimulationEvent.cpp
@@ -143,23 +143,15 @@ auto collision_time(
         return { true, t };
     }
     else if (disc > 0) {
-        auto t1 = (-b + std::sqrt(disc)) / (2 * a);
-        auto t2 = (-b - std::sqrt(disc)) / (2 * a);
+        const auto t1 = -b + std::sqrt(disc);
+        const auto t2 = -b - std::sqrt(disc);
 
-        // Handle cases where t2 is lower than t1
-        if(t2 < t1) {
-            std::swap(t1, t2);
-        }
-
-        if(t1 >= 0.0 && t1 <= 1.0) {
-            /// t1 is valid to pick, because it is in between 0 and 1
-            return { true, t1 };
-        } else if (t1 <= 0.0 && t2 >= 0.0) {
-            /// t1 is before 0 and t2 is after, but that means at t = 0.0 we are colliding
-            return { true, 0.0 };
+        if (t1 >= 0.0 && t2 >= 0.0) {
+            return { true, std::min(t1, t2) / (2 * a) };
+        } else if (std::min(t1, t2) <= 0.0 && std::max(t1, t2) >= 0.0) {
+            return { true, 0.0}
         } else {
-            /// The range [0, 1] does not intersect with [t1, t2]
-            return { false, 0.0 };
+            return { true, std::max(t1, t2) / (2 * a) };
         }
     }
     else {
--- a/environment/core/SimulationEvent.cpp
+++ b/environment/core/SimulationEvent.cpp
@@ -143,8 +143,13 @@ auto collision_time(
         return { true, t };
     }
     else if (disc > 0) {
-        const auto t1 = (-b + std::sqrt(disc)) / (2 * a);
-        const auto t2 = (-b - std::sqrt(disc)) / (2 * a);
+        auto t1 = (-b + std::sqrt(disc)) / (2 * a);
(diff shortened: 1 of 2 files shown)


Janzert:
Checking that 0 <= time <= 1 is handled by the caller.

Also I haven't looked at either the changes or the original enough to understand either, but some brief local testing seems to show that the changes make the function miss many collisions.

nomad010:
Yes, I have fixed the missed cases. I also missed the check for t in range in the caller. However, there is still a check for min(t1, t2) < 0 and max(t1, t2) > 0 where the circles are colliding initially. In that case you will say the initial impact is when the points is at max(t1, t2) instead of 0. (not that it has any material impact on the game). I will close the PR if you want.

Janzert:
Although it's not used by the engine (and even of debatable use) I think we should leave the current behavior intact of returning the collision time whenever it occurs whether in the past or far in the future.

On the other part, after looking at it a bit more I definitely still don't understand everything that is happening in the function. But if what I did pick up is correct, t1 and t2 are essentially the times at which the collision begins and ends. So if their sign matches you want the closest to 0 (i.e. the one that is closest in the past or future respectively). But when one is negative and the other positive that indicates we are in the midst of the collision and should return 0.

Currently the function only tests if both are >= 0 and returns the smallest one (as it should), but then it just lumps the other two cases together and returns the largest one. This gives an erroneous result if one is negative and the other positive. This will lead to actual incorrect results in games, although probably fairly rarely. In the case of two ships that start a turn within combat range and are moving but don't get out of range within the turn they won't trigger an attack between them (since the collision time returned by the function will be larger than 1).

I would recommend going back to the original if/else sequence, change the original else clause to an `else if (t1 <= 0 and t2 <= 0) {` and add a new else that return { true, 0.0 }. I believe this maintains the current behavior in regards to negative and times larger than 1 while also fixing the bug.

nomad010:
I will make the changes you requested.

nomad010:
I have made the changes, but I have only added a single check for where one is negative and one is positive.

environment/core/SimulationEvent.cpp
```diff
@@ -148,8 +148,9 @@ auto collision_time(
 
         if (t1 >= 0.0 && t2 >= 0.0) {
             return { true, std::min(t1, t2) / (2 * a) };
-        }
-        else {
+        } else if (std::min(t1, t2) <= 0.0 && std::max(t1, t2) >= 0.0) {
```
Janzert: This works, but it means having two unnecessary min and max calls. Also the the 'in a collision' case is almost certainly the least common so should be slightly more performant at the end of the if/else if/else chain.

I also think having the first option `if (t1 >= 0.0 && t2 >= 0.0)` and the second option `else if (t1 < 0.0 && t2 < 0.0)` makes it more clear what the three branches are actually covering.

Having said that, the code as is looks like it should work correctly.

environment/core/SimulationEvent.hpp
```diff
@@ -6,6 +6,7 @@
 #include <iostream>
 #include <unordered_set>
 #include <vector>
+#include <algorithm>
```
Janzert: This is also no longer needed.

nomad010: Oops. I will remove.

environment/core/SimulationEvent.cpp
nomad010: Technically, t2 should always be less than than t1 according to the calculations which would simplify the code, but the fact is not used elsewhere in the code so I did not either. I will make the changes and put another else if as you suggested.

Janzert: Just to be sure it's clear. There doesn't need to be a second `else if`, just make the current `else` branch the `else if` using the condition above and move the `return { true, 0.0 };` into the `else` branch.

Thanks for finding the issue and putting in the work to fix this.

nomad010:
Comments addressed.

environment/core/SimulationEvent.cpp
nomad010: I have done as you asked. Thanks for the comments and taking the time to look at PR.

harikmenon:
@Janzert : is it approved from your end?

PR #204, opened 2017-11-03
Ruby starter: Fix planets' references to docked ships

Billiam:
Fixes a string/integer mismatch for a planet's docked ships. 

Planets would have populated ship ids (so that docked ship count was accurate), but would not link these correctly to ship instances.


harikmenon:
Thanks for the contribution,

PR #212, opened 2017-11-04
Fix angle calculation

courtenay:
Angle returned from atan can be < 0, fixes this (copy solution from JS bot)


harikmenon:
Looks like a reasonable change. Thanks for the contribution.

PR #214, opened 2017-11-04
Fix misleading comments regarding BASE_PRODUCTIVITY

liviuchircu:
Similar issue to #205.


harikmenon:
Thanks for the contribution!

PR #215, opened 2017-11-05
Batch scripts now compile and run for scala REPL

RuMcClu:
also added MAX_CORRECTIONS to the constants file

### DIFF
--- a/airesources/Scala/hlt/Constants.scala
+++ b/airesources/Scala/hlt/Constants.scala
@@ -8,7 +8,7 @@ object Constants {
   val MAX_SPEED = 7
 
   /** Radius of a ship */
-  val SHIP_RADIUS = 0.5
+  val SHIP_RADIUS: Double = 0.5
 
   /** Starting health of ship, also its max */
   val MAX_SHIP_HEALTH = 255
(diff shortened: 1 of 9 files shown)


airesources/Scala/hlt/Constants.scala
```diff
@@ -34,15 +34,15 @@ object Constants {
   /** Number of turns it takes to dock a ship */
   val DOCK_TURNS = 5
 
-  /** Number of production units per turn contributed by each docked ship */
+  /** Number of turns it takes to create a ship per docked ship */
```
Janzert: Might be a merge issue but, the previous phrase was correct. This is the production units produced per docked ship per turn, not the turns needed to produce a ship per docked ship.

Snaar:
This definitely looks like bad merge. I'm afraid it would make sense to recreate this pull request with clean set of changes.

Snaar:
DId you only intend to change run_game.bat and run_game.sh? Changes to those look clean, but rest looks like bad merge.

Snaar:
Went ahead and fixed myself in fc284fc2a2ff6278449a6d24b85ffaa0f44b9963 and db94c510b9bc8c4c7bd15177ff97b1dcc87f46ef

Thanks for feedback!

PR #216, opened 2017-11-05
Ruby: Send bot name after first map update

Billiam:
Allow users to take advantage of initial 60 second timeout for map preprocessing for #151


harikmenon:
Thanks for the contribution.

PR #225, opened 2017-11-06
Elixir Starter Kit

joelbyler:
This is very much a work in process but looking to get some feedback from the halite maintainers while I continue to flesh out additional details.


Snaar:
Can you address #151 and modify MyBot to show where to use it?

Snaar:
Generally we would prefer to have almost same API as what other kits have, but we already have some other less-used languages as very basic kits so it's ok.

Snaar:
I would like to see utility to convert any angle in radians to clipped angle in degrees as expected by the engine - check https://github.com/HaliteChallenge/Halite-II/blob/master/airesources/Java/hlt/Util.java out for an example.

joelbyler:
@Snaar thanks for the feedback.  I'll work toward your suggestions in the next few commits.

joelbyler:
@Snaar I've added a hook for #151 to the starter pack and included a util module with a function to convert radians to degrees.  I agree that it would be good to try and match the APIs of other languages.  I'm glad to work towards that but am hoping to see if we can actually get an elixir bot running in the competition.  What are the next steps for making that a possibility?  Thanks so much for your help!

Snaar:
This looks good. Next step is to add support to the worker. Here is example of the work for another language: #253

I'm merging this request in and you can submit another PR for the worker changes.

PR #237, opened 2017-11-07
Updated docking language in a lot of languages

bovard:
### The problem
Docking check was wrong in all the starting languages

### The solution
Add ship radius to the calculation

### How to test
Make sure the starter bots still compile

### DIFF
--- a/airesources/C++/hlt/ship.hpp
+++ b/airesources/C++/hlt/ship.hpp
@@ -28,7 +28,7 @@ namespace hlt {
 
         /// Check if this ship is close enough to dock to the given planet.
         bool can_dock(const Planet& planet) const {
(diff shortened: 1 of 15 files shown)


Snaar:
@j-clap @lidavidm does the language make sense to you from game logic point of view?

lidavidm:
Yup, relevant code is here, and `dock_radius` is calculated in exactly this way.

https://github.com/HaliteChallenge/Halite-II/blob/e925016ba7542402ea9600e20ed62d698c48f6c9/environment/core/Entity.hpp#L181-L187

(you'd think, given that I wrote both this code and the original Python, that I'd have noticed this…)

Snaar:
Ok in it goes then!

PR #238, opened 2017-11-07
Use a git describe based version for the engine.

Janzert:
This will create version numbers of the form `<git tag>.<commits since tag>.<current commit>`. Current version looks like `1.2.72.ge9250`. When the git version is unavailable (due to either not building from source not in a repo or git itself not installed) it will use the stored version with ".src" appended.

One caveat with the approach here is that the version number will only update when a `cmake` command is run. I'm not sure if that's a problem in practice or not, it does mean that git pulls or commits after running cmake won't get reflected in the version number.

Edit: Should also mention that if the current git commit is tagged the version is simply `<git tag>`.


environment/CMakeLists.txt
```diff
@@ -27,6 +27,35 @@ include_directories(${CMAKE_SOURCE_DIR}/zstd-1.3.0/lib/legacy)
 
 get_property(dirs DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES)
 
+# Use git describe as the version for the executable, based on code from
+# http://eatmyrandom.blogspot.com/2010/06/automate-version-numbering-using-git.html
+find_package(Git)
+if( NOT GIT_FOUND AND EXISTS "${CMAKE_SOURCE_DIR}/../.git" )
```
lidavidm: Should this be `GIT_FOUND`, not `NOT GIT_FOUND`?

Janzert: Yep, sorry I just noticed that I left that in from testing the else case.

Janzert:
Ahh, I didn't realize cmake wasn't used at all on windows. In the current form it certainly won't work. The simplest fix would be to check in the generated `version.hpp`. I'll see if there's an easy way to generate it on windows though.

I planned to add the replay embedding after this was checked in.

Janzert:
Since the pure `cmake` approach wasn't going to work across all platforms anyway I decided to change the approach and get the version definition as a true build time activity. This adds a script that is run during the regular `make` command build process to update the `version.hpp` header if needed.

At this point windows should build without a problem but will not update the version from whatever is checked in.

Janzert:
It would also be great if someone could test this on macOS. I tried to keep everything in `set_version.sh` at a posix level of compatibility, but I don't have a mac around to test on and won't be surprised if I messed it up.

environment/version.hpp
```diff
@@ -0,0 +1,2 @@
+// Define the version of the executable
```
Snaar: Let's add comment here on how to regenerate/update the file.

Snaar:
For mac it looks reasonable enough to just merge in and then I'll check it right after.

environment/version.hpp
Snaar: Oh, was too slow. Looks good.

Snaar:
For completeness: mac turned out to have mandatory argument for -i on sed. Pushed fix out directly.

PR #240, opened 2017-11-08
Fix typo: should be WORKING_DIRECTORY

wizzup:
reference : https://cmake.org/cmake/help/v3.2/command/add_custom_target.html

### DIFF
--- a/environment/CMakeLists.txt
+++ b/environment/CMakeLists.txt
@@ -29,7 +29,7 @@ get_property(dirs DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTO
 
 add_custom_target(VERSION_CHECK
     ./set_version.sh
-    WORKING DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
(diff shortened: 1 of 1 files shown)


environment/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ get_property(dirs DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTO
 
 add_custom_target(VERSION_CHECK
     ./set_version.sh
-    WORKING DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
```
Snaar: Good find. Though this begs question, is this variable even needed? Wouldn't it default to current source dir anyway? @Janzert?

wizzup: I found the typo because my build was failed to find `set_version.sh` so I guess you might want to keep it.

I am running NixOS and I [modify](https://github.com/wizzup/Halite-II/tree/nix/environment) some code to be able to build on my system.

Snaar: O, didn't realize build failed. I'll merge then.

Issue #226, opened 2017-11-06
Create new language kit guidelines


Snaar:
Writing down my thoughts here for now.

Basic kit:
- default bot name different from default names of other languages
- has logging that creates/overwrites log file with ${id}-${name}.log as file name
- reads in the map and all the entities on it
- makes it obvious how to take advantage of pre-game 60s timeout (#151)
- utility to convert radians into clipped integer degrees as expected by the halite engine - example https://github.com/HaliteChallenge/Halite-II/blob/master/airesources/Java/hlt/Util.java
- the magnitude component of thrust command should be rounded down to prevent ship from crashing into things

Full kit:
- all of the functions on website documentation available

mikemeier:
@Snaar done

harikmenon:
Closing this out since this is the end of the competition

PR #245, opened 2017-11-08
Provide PHP Starter Kit

mikemeier:


Snaar:
I'll take a look at this in detail later in the day. For now, can you check if https://github.com/HaliteChallenge/Halite-II/issues/226 make sense and are addressed?

Snaar:
Based on initial inspection, you do need to address some of the points, such as at least the log file name and the initial 60s pre-game period.

airesources/PHP/MyBot.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+require_once __DIR__.'/src/bootstrap.php';
+
+$connection = new Connection();
+$game = new Game('Terminator', $connection);
+$turn = 0;
+
+//60s time for game initialization;
```
Snaar: I think I'm missing something. Is map actually parsed at this point? I only see it being parsed in updateMap() function but it hasn't been called here yet. I.e. how do you use the initial map here?

mikemeier: @Snaar I thought as soon as map is parsed, the 60s are not allowed anymore. So when exactly is it allowed to wait 60s?

Snaar: It's actually as soon as you send your bot name over is when the 60s timer stops. What happens is you get pre-game map first, along with map sizes and your player id. Here is where you can take 60s. Then you send your bot name over and then the game will send you the first turn map - which actually would look just like the pre-game map.

Check out how some other kits are doing it. #151 has the list of kits that are doing it right currently.

Snaar: (This is the only remaining issue before I'd merge it in btw. Rest looks good.)

Snaar:
Actually I'm just going to merge it in as-is. But the #151 will still need to be addressed for PHP kit.

PR #251, opened 2017-11-08
Portable random number generation

lidavidm:
Pulls the Clang libcxx implementation of `uniform_int_distribution` and `uniform_real_distribution` (MIT license) into the engine itself, hopefully giving us deterministic random number generation across platforms.

### DIFF
--- a/environment/core/util/distributions.hpp
+++ b/environment/core/util/distributions.hpp
@@ -1,0 +1,375 @@
+//===----------------------------------------------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is dual licensed under the MIT and the University of Illinois Open
+// Source Licenses. See below for details.
+//
+//===----------------------------------------------------------------------===//
+
+// Derived from Clang libcxx 7e73ea8.
+
+// ==============================================================================
+// libc++ License
+// ==============================================================================
+
+// The libc++ library is dual licensed under both the University of Illinois
+// "BSD-Like" license and the MIT license.  As a user of this code you may choose
+// to use it under either license.  As a contributor, you agree to allow your code
+// to be used under both.
+
+// Full text of the relevant licenses is included below.
+
+// ==============================================================================
+
+// University of Illinois/NCSA
+// Open Source License
+
+// Copyright (c) 2009-2017 by the contributors listed in CREDITS.TXT
+
+// All rights reserved.
+
+// Developed by:
+
+//     LLVM Team
+
+//     University of Illinois at Urbana-Champaign
+
+//     http://llvm.org
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy of
+// this software and associated documentation files (the "Software"), to deal with
+// the Software without restriction, including without limitation the rights to
+// use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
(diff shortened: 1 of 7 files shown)


environment/core/util/distributions.hpp
```diff
@@ -0,0 +1,474 @@
...
+        // Search from LSB to MSB for first set bit.
+        // Returns zero if no set bit is found.
+        if (_BitScanReverse(&where, mask))
+            return 31 - where;
+        return 32; // Undefined Behavior.
+#endif
+    }
+
+    static unsigned long __clz(unsigned long __x) {
+#ifndef _LIBCPP_COMPILER_MSVC
```
Snaar: Can you add this define to the make.bat, make2017.bat, make2017_32.bat? I promise I'll turn them into single bat file soon, been working on cleaning that up :P

environment/core/util/distributions.hpp
```diff
@@ -0,0 +1,474 @@
...
+        return static_cast<result_type>(__u + __p.a());
+    }
+
+    template<class _RealType, std::size_t __bits, class _URNG>
+    _RealType
+    generate_canonical(_URNG& __g)
+    {
+        const std::size_t _Dt = std::numeric_limits<_RealType>::digits;
+        const std::size_t __b = _Dt < __bits ? _Dt : __bits;
+#ifdef _LIBCPP_CXX03_LANG
```
Snaar: This looks safe for me to leave as is undefined anywhere. Just fyi for completeness.

Janzert:
This currently misses the distribution at line 196 of main.cpp where map size is set. It also doesn't update the asteroid cluster generator, but that is unused at this point so probably doesn't matter.

environment/core/util/distributions.hpp
```diff
@@ -0,0 +1,474 @@
...
+    }
+
+    static unsigned long long __clz(unsigned long long __x) {
+#ifndef _LIBCPP_COMPILER_MSVC
+        return static_cast<unsigned long long>(__builtin_clzll(__x));
+#else
+        unsigned long where;
+        // BitScanReverse scans from MSB to LSB for first set bit.
+        // Returns 0 if no set bit is found.
+#if defined(_LIBCPP_HAS_BITSCAN64)
```
Snaar: Ok to leave undefined. We do support building for 32 bit windows, which lacks _BitScanReverse64. Don't care about performance implications here.

environment/core/util/distributions.hpp
lidavidm: Done, thanks for pointing that out.

environment/core/util/distributions.hpp
lidavidm: I went ahead and took out that branch, since we don't need it anyways.

lidavidm:
@Janzert thanks for pointing that out, updated that as well.

Snaar:
Ok, had to do 7834a13c5e1e9dcbdfb475a11487d34ca4fe9695 and 8eb11f6c5f9072ba3117bfb3e381416ccbfdef70 to make it compile and work on windows. It produces identical results now for same seed as linux version.

Snaar:
It generates quite some warnings on windows now - I'll clean it up later.

PR #252, opened 2017-11-08
Updated Julia starter kit to be more "julian"

shorty66:
I made a quick review of the starter kit and removed some unnesseccary inner constructors, replaced @sprintf with string interpolation in many places and tidied up here and there.

This currently works on my local machine. My current bot runs on a very similar version online, but i didn't run any online tests with this PR.

### DIFF
--- a/airesources/Julia/hlt/entity.jl
+++ b/airesources/Julia/hlt/entity.jl
@@ -121,13 +121,6 @@ function Ship(owner_id::String, tokens::Vector{String})
 end
 
 """
-    isdocked(ship)
-
-Test if current ship is undocked. Returns `true` if ship is `UNDOCKED`, `false` otherwise.
-"""
-isdocked(ship::Ship) = ship.docked != UNDOCKED
-
-"""
     can_dock(ship, planet)
 
 Determine whether a ship is close enough to planet so it can dock.
@@ -136,32 +129,12 @@ function can_dock(ship::Ship, planet::Planet)
     calculate_distance_between(ship, planet) <= radius(planet) + Constants.DOCK_RADIUS + Constants.SHIP_RADIUS
 end
 
-"""
-    thrust(ship, speed, angle)
-    
-Generate a command to accelerate this ship.
-"""
-function thrust(ship::Ship, speed, angle)
-    @sprintf("t %s %s %s", ship.id, floor(Int, speed), round(Int, angle))
-end
 
-"""
-    dock(ship, planet)
+isdocked(ship::Ship) = ship.docked != UNDOCKED
+dock(ship::Ship, planet::Planet) = "d $(ship.id) $(planet.id)"
+undock(ship::Ship) = "u $(ship.id)"
 
-Generate a command to dock to a planet.
-"""
-function dock(ship::Ship, planet::Planet)
-    @sprintf("d %s %s", ship.id, planet.id)
-end
-
-"""
-    undock(ship)
-
-Generate a command to undock from the current planet.
-"""
-function undock(ship::Ship)
-    @sprintf("u %s", ship.id)
-end
+thrust(ship::Ship, speed, angle) = "t $(ship.id) $(floor(Int, speed)) $(round(Int, angle))"
 
 function nearest_unoccupied_planet(ship::Ship, planets)
     d = Inf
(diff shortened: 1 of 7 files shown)


airesources/Julia/hlt/entity.jl
```diff
@@ -103,66 +103,38 @@ struct Ship <: Entity
...
-
-Test if current ship is undocked. Returns `true` if ship is `UNDOCKED`, `false` otherwise.
-"""
-isdocked(ship::Ship) = ship.docked != UNDOCKED
+function Ship(owner_id::String, tokens::Vector{String})
+    id = shift!(tokens)
+    x = parse(Float64, shift!(tokens))
+    y = parse(Float64, shift!(tokens))
+    hp = parse(Int, shift!(tokens))
+    vel_x = parse(Float64, shift!(tokens))
```
Snaar: Can you actually drop the vel_x and vel_y? They are deprecated so having those variables in the kit might be confusing to people.

Snaar: (I.e. still read them since they are still part of the protocol with game engine, but don't store the values in any variables.)

shorty66: yeah of course i will. I just edited the things that seemed obviously strange to me.

airesources/Julia/hlt/game.jl
```diff
@@ -1,93 +1,53 @@
...
+start_game(game::Game) = print(game.botname, "\n")
 
 """
     setup_logger(name, id)
 
 Set up and truncate the log. Log name is "{name}-{id}.log".
 """
 function setup_logger(name, id)
-    add_handler(logger, DefaultHandler(open(@sprintf("%s-%s.log", name, id), "w"), DefaultFormatter("[{date} | {level} | {name}]: {msg}"))) 
+    add_handler(logger, DefaultHandler(open("$name-$id.log", "w"), DefaultFormatter("[{date} | {level} | {name}]: {msg}")))
```
Snaar: Can you do $id-$name instead? That's the convention with all other kits.

airesources/Julia/hlt/halite.jl
```diff
@@ -1,44 +1,46 @@
...
-    #: Number of production units per turn contributed by each docked ship
-    const BASE_PRODUCTIVITY = 6
+    DOCK_TURNS::Int
+    #: Number of turns it takes to create a ship per docked ship
+    BASE_PRODUCTIVITY::Int
     #: Distance from the planets edge at which new ships are created
-    const SPAWN_RADIUS = 2.0
+    SPAWN_RADIUS::Float64
 end
+Constants = CONST(7, 0.5, 255, 255, 1, 5.0, 64, 10.0, 4.0, 4, 6, 2.0)
```
Snaar: This kinda sucks. Let's say you want to change the MAX_SHIP_HEALTH - now you have to guess/reverse-engineer which one of the 255 it is.

airesources/Julia/hlt/entity.jl
Snaar: Yup, just taking it as opportunity to clean it up a bit more.

airesources/Julia/hlt/halite.jl
shorty66: yeah \, i figure it would be best to jsut declare the constants without hte struct or module. All Uppercase should indicate clear enough that its a constant.

airesources/Julia/hlt/halite.jl
```diff
@@ -15,32 +19,30 @@ include("utils.jl")
...
+const WEAPON_RADIUS = 5.0
+#: Weapon damage
+const WEAPON_DAMAGE = 64
+#: Radius in which explosions affect other entities
+const EXPLOSION_RADIUS = 10.0
+#: Distance from the edge of the planet at which ships can try to dock
+const DOCK_RADIUS = 4.0
+#: Number of turns it takes to dock a ship
+const DOCK_TURNS = 4
+#: Number of turns it takes to create a ship per docked ship
```
liviuchircu: This is outdated, see 52da46ed7

Snaar:
Follow up comment: you should probably be more explicit if angle is in deg or rad. For example:

`angle_between(e1::Entity, e2::Entity) = rad2deg(atan2(e2.y - e1.y, e2.x - e1.x)) % 360`

Maybe that should be called "angle_between_in_deg"? Also maybe this needs to be rounded to an integer first before % 360?

Issue #135, opened 2017-10-29
Collision occurs the turn after


shummie:
See attached replay:
[replay-20171028-214016-0500--1898579245-312-208-90311.zip](https://github.com/HaliteChallenge/Halite-II/files/1424759/replay-20171028-214016-0500--1898579245-312-208-90311.zip)

On turn 32: ship id 11 is at position 128.79/90.74 and moves. On the web replay viewer, the ship is there. Now, when you advance 1 turn, the map object properties shows that the ship position for 11 is now at 133.79/91. However, if you look at the ship in the replay viewer, it doesn't actually move.

Advance 1 more frame, and then it looks like the ship crashes into Ship 9 which was docking onto the planet.

The distances between the two ships are 0.999974, so the ships should crash, but it's weird that 1) the replay viewer doesn't update the position of Ship 11 between frames 32 and 33, and then 2) the collision event occurs after the ship has already moved and is issuing a dock command when it should have probably crashed on frame 33 instead of 34.

Might be related to the below:
#114 
#129 

Note, this was tested locally, and i don't have the most up to date binaries as of 10/28/17. I can test with the same seed if a new binary is available to see if #129 fixes this.

lidavidm:
Hmm, what I see in the replay is that the distance between ships 9 and 11 on frame 33 is 1.0000000000004252. The visualizer rounds the positions to be a little more friendly, but it can be misleading. The ships not moving is definitely a visualizer bug, though.

lidavidm:
Ah wait, I just saw the discussion on Discord. I'm going to take a closer look once I find a good block of time…

lidavidm:
@shummie , as far as I can tell, what's happening is:

- At the beginning of turn 33, ships 9 and 11 are alive, and so commands get issued. Ship 11 issues a dock command; ship 9 is already docking.
- They collide right away, and are destroyed on turn 33. (I'm basing this on the `events` field in the frame)
- The distance between them is slightly greater than 1 when they collide, and both should be stationary - looking into why that happens; it may be due to the residual velocity, if that got rounded down when writing the replay.

The visualizer doesn't appear to process the event until the end of the frame, for some reason. There is some wonkiness, especially when advancing frame by frame.

shummie:
Yeah, your 3rd bullet point is the weird one. I downloaded the latest binary to see if the newest build which fixes some other things w/ velocity fixes it but I can still reproduce the same issue. Thanks for continuing to take a look at it! 

[...]

PR #263, opened 2017-11-09
Increase SERIALIZATION_PRECISION to 6

fohristiwhirl:
I don't see any real downside to this. This number only determines the precision sent to the clients, it doesn't affect replays.

We could even go higher e.g. 8 or 10. But I suppose it's nice to be able to look at the comms visually if one wants to understand the format (e.g. [I did so here](https://forums.halite.io/t/halite-comms-protocol-explained/411/1)). So perhaps 6 is fine.

### DIFF
--- a/environment/networking/Networking.hpp
+++ b/environment/networking/Networking.hpp
@@ -32,7 +32,7 @@ extern bool quiet_output;
 /**
  * How many digits of precision to send to the client.
  */
-constexpr auto SERIALIZATION_PRECISION = 4;
+constexpr auto SERIALIZATION_PRECISION = 6;
 
 constexpr auto INIT_TIME_LIMIT = std::chrono::seconds{60};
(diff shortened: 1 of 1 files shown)


Snaar:
I wonder if there is any reason not to do full precision?

fohristiwhirl:
Want to go to 8 or 10? The only issues are:

* Some insanely slow language might have trouble parsing the extra characters? (Unlikely.)
* The format is less human-readable.

lidavidm:
Yeah, I don't see an issue with it. The main thing was not wanting bots to get too tied up in floating point details (still need to fix #135, for instance)

Snaar:
One vote here for full precision output.

fohristiwhirl:
Well. Can I leave it to TS people to decide a value and apply it?

lidavidm:
See #270 for full precision then

Snaar:
Just merged one with full precision in, so this is not needed.

PR #253, opened 2017-11-09
Added swift support to worker

andybest:
Hey, I believe that this should be all that is necessary to add swift support to the workers?

PR #266, opened 2017-11-10
Attempt at adding elixir support for workers

joelbyler:
Work in process, any advice you can give for testing this out would be super helpful :)

As far as how I build this locally, this is the process I use:

## Build the bot
```
cd airesources/Elixir/elixirbot
mix deps.get
mix escript.build
```

## Run a two bot match
```
./halite -d "240 160" "./elixirbot/elixirbot" "./elixirbot/elixirbot"
```

I'm having trouble understanding how to map those commands to the changes I need to make to the repo, @Snaar pointed me at #253 so I'll be looking over that PR but any advice folks can give along the way for this will be super helpful.  Thanks!


apiserver/worker/compiler.py
```diff
@@ -408,7 +408,13 @@ def compile(self, bot_dir, globs, errors, timelimit):
     ),
     Language("Dart", BOT +".dart", "MyBot.dart",
         "dart MyBot.dart",
-        [], [(["*.dart"], ChmodCompiler("Dart"))]),
+        [], [(["*.dart"], ChmodCompiler("Dart"))]
+    ),
+    Language("Elixir", BOT +".ex", "MyBot.ex",
+        "./MyBot",
```
lidavidm: This setup requires that the Elixir bot always have a shebang line; is that intentional? Otherwise it might make more sense to have the command executed be `elixir MyBot` or something like that.

joelbyler: @lidavidm good question, currently the way the elixir starter kit works is that it uses `escript` to generate an elixir cli interface as its own stand-along executable.  So we can just run it as `./Mybot`as shown in this example [here](https://github.com/HaliteChallenge/Halite-II/blob/master/airesources/Elixir/run_game.sh#L3) but maybe that PR was merged in too hastily.  I'm glad to rework that starter kit to depend on running the bot via the `elixir MyBot.exs` if that's preferable.  Thanks for the feedback and I look forward to your response.

lidavidm: Ah, I didn't know that. In that case, this looks good - I'll approve it. I'll try running that bot later and just make sure it's all working. One note though, there is a merge conflict (another package got added in the meantime), if you could address that.

joelbyler:
Thanks @lidavidm, I merged in the upstream changes so those conflicts should be resolved now.

apiserver/worker/compiler.py
lidavidm: So @joelbyler , is the Elixir starter kit able to be zipped up and submitted as-is? There isn't any `MyBot.ex` - or should `elixirbot` be renamed to `MyBot.ex` first?

lidavidm: (The `Language` part specifies that for Elixir, we look for a `MyBot.ex` file, run `chmod` on it, then look for a `BOT + '.ex' == 'MyBot.ex'` file to take as the actual bot. It then writes `./MyBot` to a bash script used to start the bot, so right now a few things don't line up.)

lidavidm:
Also just noticed - the `ChmodCompiler` doesn't actually work! It sets the permissions to 644 (probably want 755), for whatever reason…the other languages that use it all explicitly run the bot, so they didn't run into it. I will fix the `ChmodCompiler`.

apiserver/worker/compiler.py
lidavidm: Anyways, once the bug with `ChmodCompiler` is fixed, it compiles fine if `elixirbot` is renamed `MyBot.ex`.

joelbyler: @lidavidm interesting, this has helped me to understand a bit more about how this works now.  I'm going to try and make a new revision to see if I can make this more straight forward.  Thanks a bunch!

joelbyler:
@lidavidm this last commit probably seems drastic but it brings the starter kit closer to what a typical elixir dev would expect from a project structure perspective and I'm hoping it makes the worker configuration a little more clean for running elixir bots.  How does this look to you?

lidavidm:
That is…indeed quite drastic, but now it seems runnable. I'll give it a test.

apiserver/worker/compiler.py
```diff
@@ -302,6 +302,10 @@ def compile(self, bot_dir, globs, errors, timelimit):
     "D": [
         ["dmd", "-O", "-inline", "-release", "-noboundscheck", "-of" + BOT],
     ],
+    "Elixir": [
+        ["mix", "deps.get"]
+        ["mix", "escript.build"],
+    ],
```
lidavidm: Hmm, this is a little more complicated now. One, it requires interactive input:
```
$ mix deps.get --force
Could not find Hex, which is needed to build dependency :logger_file_backend
Shall I install Hex? (if running non-interactively, use "mix local.hex --force") [Yn]
```

If I do that, everything else seems fine. Should this just be added to the list of commands?

Does `mix` install dependencies locally (ala `node_modules` in NPM) or does it install them in some global or user directory? If the latter, that would cause problems (they won't get included in the zip).

Also, this is rather confusing, but the compiler script won't actually run these commands automatically (yeah, I know…) In the `Language` definition for Elixir below, in the same list as the `ChmodCompiler` command, you'll have to specify that you want to invoke these commands, like D does:

```
(["*.d"], ExternalCompiler(comp_args["D"][0]))
```

The first part of the tuple is a list of globs to pass to the compiler specified; the second part is the actual command to execute (which refers back to the `comp_args` here). So the final list (below) would look something like 

```
[
    ([], ErrorFilterCompiler(com_args["Elixir"][0])),
    ([], ErrorFilterCompiler(com_args["Elixir"][1])),
    (["*.ex"], ChmodCompiler("Elixir")),
]
```

If you want, I can tweak it and submit a PR to your branch.

Finally, there's still a mismatch in terms of the file it looks for (as the 'bot' file) and the file it runs. Probably we should have

```
Language("Elixir", BOT, "mix.exs",
```
which specifies that we detect that the bot is Elixir by looking for a file `mix.exs`, which after we run the compiler commands, produces a file called `MyBot`.

airesources/Elixir/run_game.bat
```diff
@@ -1 +1,3 @@
+mix deps.get
+mix escript.build
 halite.exe -d "240 160" "./elixirbot/elixirbot" "./elixirbot/elixirbot"
```
BobbyMcWho: This should also be changed to `./MyBot`, correct?

admin/setup_worker_image.sh
```diff
@@ -29,11 +29,15 @@ sudo sh -c 'curl https://storage.googleapis.com/download.dartlang.org/linux/debi
...
 
+## Add for Elixir Support
+wget https://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb && sudo dpkg -i erlang-solutions_1.0_all.deb
+
 sudo apt-get update
 sudo apt-get -y upgrade
 
 ## List the packages to install for running bots.
-PACKAGES="build-essential gcc g++ python3 python3.6 python3-pip git ocaml openjdk-8-jdk php ruby scala nodejs mono-complete dotnet-dev-1.1.0 libgeos-dev tcl8.5 mit-scheme racket octave luajit lua5.2 ghc erlang-base-hipe coffeescript dart fp-compiler sbcl dmd-bin mono-vbnc gnat-6 cmake python3.6-dev python-numpy cython clang libicu-dev"
+PACKAGES="build-essential gcc g++ python3 python3.6 python3-pip git ocaml openjdk-8-jdk php ruby scala nodejs mono-complete dotnet-dev-1.1.0 libgeos-dev tcl8.5 mit-scheme racket octave luajit lua5.2 ghc erlang-base-hipe coffeescript dart fp-compiler sbcl dmd-bin mono-vbnc gnat-6 cmake python3.6-dev python-numpy cython clang libicu-dev esl-erlang elixir"
```
BobbyMcWho: I notice that erlang-base-hipe is already present here @joelbyler , from my limited research, it seems like this could cause issues when installing esl-erlang afterwards?

BobbyMcWho: It seems like esl-erlang includes all the packages and erlang-base-hipe could be removed

joelbyler: Thanks for checking this out @BobbyMcWho, admittedly I simply just pulled those from the elixir guide but that makes a lot of sense and should clean this up.  Nice catch, thanks!

joelbyler:
@lidavidm 
Thanks for checking this out so closely, I've updated the PR to hopefully work for all of those things you highlighted.  I'll try and sum the changes up here:
* I changed the command to pipe `yes` to bypass the interaction.  I'm not sure there is actually any documented support for the `--force` option but in the past I believe folks have used this alternative approach to get around that prompt.
* I believe the dependencies will be stored in the `deps` folder within the Elixir directory by default.
* The way `ErrorFilterCompiler` and those arguments that get passed to `Language` work are much more clear now, thanks for explaining that.

joelbyler:
@lidavidm one followup question, because the second mix command actually generates the executable, `MyBot` then I don't think a `chmod` is required, but perhaps this needs to run as a different user or something?  I think I'm seeing other languages that don't have anything to change permissions though.  Do you have any thoughts there?

lidavidm:
If `mix` generates it as executable then there's no need for a chmod. I'll give it a run later.

lidavidm:
Thanks for updating this so much - it's definitely not the best-documented part of the system.

apiserver/worker/compiler.py
```diff
@@ -302,6 +302,10 @@ def compile(self, bot_dir, globs, errors, timelimit):
     "D": [
         ["dmd", "-O", "-inline", "-release", "-noboundscheck", "-of" + BOT],
     ],
+    "Elixir": [
+        ["yes", "|", "mix", "deps.get"]
```
lidavidm: Small typo - needs a comma after the list

apiserver/worker/compiler.py
```diff
@@ -412,7 +416,16 @@ def compile(self, bot_dir, globs, errors, timelimit):
...
     Language("Dart", BOT +".dart", "MyBot.dart",
         "dart MyBot.dart",
-        [], [(["*.dart"], ChmodCompiler("Dart"))]),
+        [], [(["*.dart"], ChmodCompiler("Dart"))]
+    ),
+    Language("Elixir", BOT, "mix.exs",
+        "./MyBot",
+        [],
+        [
+            ([], ErrorFilterCompiler(com_args["Elixir"][0])),
```
lidavidm: Small typo (my bad) - should be `comp_args` not `com_args` here and below

joelbyler: Cool, I'll fix these and squash the commits so this doesn't look so noisy when someone looks thru history.

lidavidm:
Sorry, passed over this, looks good now.

Snaar:
Actually just noticed https://github.com/HaliteChallenge/Halite-II/pull/266/files#diff-46ee9149e8bc859a7547cc20f2c4c9bf - is this the bot binary being committed? If i'm not misreading, we should not commit binary.

Snaar:
Yeah ok, there is binary in github already right now. Need to remove that.

lidavidm:
@harikmenon new images for this built, by the way (also checked that we have Pandas etc all installed)

Snaar:
Lemme just merge it in and remove the file myself for expediency.

Snaar:
Ok there is a bit more to clean up. Will need just a bit of time.

Snaar:
Ok, stuff is clean now. However I believe there is some bug in the kit, since bots don't die natural death, they are currently killed by halite itself.

BobbyMcWho:
@snaar if it's the same logic as the original elixir starter package, the bots move at random and eventually run into planets and die

Snaar:
I haven't ran the original so wouldn't know. I just know that current one doesn't terminate correctly for whatever reason - i.e. it's not just gameplay level issue, it's some kind of fundamental issue with the kit right now.

Snaar:
I.e. I see this at the end of a game:
```
Turn 295
Map seed was 943919007
Opening a file at ./replay-20171112-214223-0500--943919007-240-160-1510540941.hlt
Player #0, Alchemist, came in rank #1 and was last alive on frame #295, producing 0 ships and dealing 0 damage!
Player #1, Alchemist, came in rank #2 and was last alive on frame #295, producing 0 ships and dealing 0 damage!
Bot 0 was killed.
Here is the rest of its output (if any):
t 1 5 185

--- End bot output ---
Bot 1 was killed.
Here is the rest of its output (if any):
t 5 6 282

--- End bot output ---
```

While it should look something like this using C++ bot as an example:
```
Turn 27
Map seed was 2720188280
Opening a file at ./replay-20171112-221157-0500--2720188280-240-160-1510542717.hlt
Player #0, IvanTheTerrible, came in rank #2 and was last alive on frame #26, producing 0 ships and dealing 0 damage!
Player #1, IvanTheTerrible, came in rank #1 and was last alive on frame #27, producing 1 ships and dealing 256 damage!
```

lidavidm:
Hmm, I've seen that happen with various bots. All the bots are killed at the end of the game (in the Halite destructor), so if there happened to be some extra output the bot had written, it'll get printed out.

Snaar:
Elixir one consistently does it though. Most of the time it seems it's just empty but still present:
```
Turn 26
Map seed was 1181036402
Opening a file at ./replay-20171112-214618-0500--1181036402-240-160-1510541178.hlt
Player #0, Alchemist, came in rank #1 and was last alive on frame #26, producing 0 ships and dealing 0 damage!
Player #1, Alchemist, came in rank #2 and was last alive on frame #25, producing 0 ships and dealing 0 damage!
Bot 0 was killed.
Here is the rest of its output (if any):

--- End bot output ---
Bot 1 was killed.
Here is the rest of its output (if any):

--- End bot output ---
```

joelbyler:
Yeah, Bobby's correct in that the starter kit doesn't have any logic in
place to actually try and win the game.  I've tried to run the script
locally myself and see the same output as @snaar.  I'm sure this isn't
something in the starter kit outputting this but perhaps the elixir code is
ignoring a dead bot state or something?  I can look thru that, I was
planning to improve the starter bot logic to at least try and dock if there
is an uninhabited planet still available.  While I'm at it I'll look thru
the other examples to see if there's something that they're doing that I
missed in the original starter kit PR.  Thanks for helping out everyone!

joelbyler:
Also, @BobbyMcWho let me know if you're interested in working together on
improving the elixir starter kit or maybe if you've already working on it?

BobbyMcWho:
@joelbyler yeah I definitely am interested, I was actually going to use this to try and learn some elixir when I saw it was an option, but then I saw the backend wasn't fleshed out yet, so I haven't messed with it yet.

PR #267, opened 2017-11-10
Added swift starter kit

andybest:


andybest:
Looks like the CI failure was something to do with NPM and not to do with the starter kit?

Snaar:
Yeah if only one of them fails it's usually environment being flaky.

airesources/Swift/Sources/hlt/Navigation.swift
```diff
@@ -0,0 +1,57 @@
...
+    
+    let thrust: Int
+    
+    if distance < Float(maxThrust) {
+        thrust = Int(distance)
+    } else {
+        thrust = maxThrust
+    }
+    
+    let angleDeg = angleRad.radiansToDegrees.truncatingRemainder(dividingBy: 360)
```
Snaar: Can you round angle to nearest int here instead? I would imagine swift rounds down when you cast float to int directly.

Snaar: You also should consider moving this into Util.radiansToDegreesClipped or something similarly named. I imagine this will be useful to people writing bots in other contexts.

airesources/Swift/Sources/hlt/Constants.swift
```diff
@@ -0,0 +1,21 @@
...
+    public static let MaxShipHealth = 255
+    public static let BaseShipHealth = 255
+    public static let WeaponCooldown = 1
+    public static let WeaponRadius = 0.5
+    public static let WeaponDamage = 64
+    public static let ExplosionRadius: Float = 10.0
+    public static let DockRadius: Float = 4.0
+    public static let DockTurns = 5
+    public static let BaseProductivity = 6
+    public static let SpawnRadius = 2.0
```
Snaar: This one is the only one missing explicit Float - other *Radius variables have it.

Issue #247, opened 2017-11-08
Challenge player support


harikmenon:
How difficult would it be for us to write an api to challenge a player.

E.g 

1.  I can select 1 or 3 other plays and challenge them
2. We run about 30 games (max 3 challenges/day) across a combination of maps and player count
3. These matches will not affect rating in any way
4. user can see the status of the challenge (not started, started, completed)

Also:
- [x] Mu/sigma/rank to game_participant?
- [x] Field to track whether game is ranked

lidavidm:
We would need a table to store the challenges, some fields on the player table to keep track of the most recent challenge/number of challenges/day, some changes in the matchmaking code to actually issue the challenge, and changes in the worker & coordinator to keep track of the fact that we are not updating rating.

harikmenon:
Makes sense.

An alternate thing would be to have as user_challenge table that stores user to challenge id and date of creation and status, Challenges table with challenge id and participants id and updates to the game table to have a challenge id.

Would that work better, how much time would you estimate would it take to wire this up. We are seeing a lot of feedback on this front and we wanted to see if we can build this feature to enhance our growth rate.

lidavidm:
Yeah, that sounds good. I'll take a look tonight after I test my fix for #241

lidavidm:
@harikmenon Are challenges publicly visible, or only to the involved players? When matchmaking, we issue special games involving only the affected players, right?

harikmenon:
Publicity visible and yes

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

lidavidm:
One more thing - @harikmenon how should we handle ties? (Each bot gets points based on their rank in the game; highest points wins, but there can be ties). Also, if they challenge 3 people, but 2-player games are played, there might be an unequal number of games between the participants.

harikmenon:
cant we use our tiebraker logic for this? I think its fine, its a best faith effort on our side to run 30 games with as many variations as we can

lidavidm:
So the issue is breaking ties after they've played the 30 games, I guess we can parse the replays and aggregate that information then.

Issue #233, opened 2017-11-07
Challenge a friend: Join & Play


julskast:
1.5 Join - New players only
- New player gets email invite
- New player visits the site and sees instructions (create a new learn page?)
- New player makes an account (what if different email?) - does challenger get notified?
- New user goes through onboarding & submits a bot - does challenger get notified?
- Does new user get to say when they're ready to accept the challenge? E.g. if you challenge me, chances are you've been playing already. I'd like to get my bot good before I accept your challenge.

2. Play
- Does challengee need to accept a challenge? What happens for top players - will they get challenged too often? Or does challengee even get notified? Maybe players can turn off challenge notifications separately via sendgrid as a new unsub group.
- Challenger plays challenged (backend)
- Challenger & challenged learn they've played (notification? Emails?)

Sample copy:
"Hi Julskast, we're letting you know that you've started playing Harikmenon as part of a player-to-player challenge. To see the results of the challenge you can go to the "analyze" section of your user profile to see stats, or watch a video of the two of you"

Janzert:
Just a note, these games should not cause rating updates. Allowing rated games with self selected opponents is an invitation for rating manipulation.

PR #271, opened 2017-11-10
Challenges

lidavidm:
Plus some other fixes

Implements #247 

New API URLs: `/v1/api/user/<user_id>/challenge` (GET, POST), `/v1/api/user/<user_id>/challenge/<challenge_id>` (GET)

Let me know if more are needed.


harikmenon:
@lidavidm have you created the tables yet?

lidavidm:
No, not yet.

lidavidm:
I can go ahead and create them - is this what you expected in terms of functionality @harikmenon ? Looking at the other issues, this doesn't support challenging/inviting players who haven't yet registered. We could add that later though. The other thing is that this diverts up to 5% of games to challenges, dunno if that's too high or not enough.

apiserver/alembic/versions/201711081734_49be2190c22d_add_challenges.py
```diff
@@ -0,0 +1,100 @@
...
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import mysql
+
+
+# revision identifiers, used by Alembic.
+revision = '49be2190c22d'
+down_revision = '33de9025cc63'
```
Janzert: What is in the repository for alembic and sql schema doesn't currently match production reality, I'll create a pull request shortly to get the repo back in line with reality again. At that point though this down_revision will probably need to be changed to whatever is actually the head revision at that point.

apiserver/apiserver/coordinator/matchmaking.py
```diff
@@ -129,9 +135,114 @@ def serve_game_task(conn, has_gpu=False):
...
+def find_challenge(conn, has_gpu=False):
+    """
+    Find a set of players in a challenge.
+    """
+    # Reset any stuck challenges
+    reset_challenges(conn)
+
+    challenge = conn.execute(
+        model.challenges.select(
+            model.challenges.c.status == model.ChallengeStatus.CREATED.value
```
Janzert: This query should incorporate getting 'stuck' challenges, something like:

    challenge = conn.execute(
        model.challenges.select(
            model.challenges.c.status == model.ChallengeStatus.CREATED.value
            | (model.challenges.c.status == model.ChallengeStatus.PLAYING_GAME.value
               & model.challenges.c.most_recent_game_task <
                 datetime.datetime.now() - datetime.timedelta(minutes=30))
        ).order_by(
            model.challenges.c.most_recent_game_task.asc()
        )
    ).first()

apiserver/apiserver/coordinator/matchmaking.py
```diff
@@ -129,9 +135,114 @@ def serve_game_task(conn, has_gpu=False):
...
+
+    selected_bots = []
+    candidate_bots = []
+    for bot in bots:
+        if bot["user_id"] == challenge["issuer"]:
+            selected_bots.append(bot)
+        else:
+            candidate_bots.append(bot)
+
+    if random.random() < 0.5 and len(candidate_bots) == 3:
```
Janzert: It seems like this is assuming the number of challenge participants will either be 2 or 4. This can be easily modified to handle an arbitrary number though.

    if random.random() < 0.5 and len(candidate_bots) >= 3:
        selected_bots.extend(random.shuffle(candidate_bots)[:3])

apiserver/alembic/versions/201710031816_7f0054256cf5_initial_database_schema.py
```diff
@@ -360,7 +360,7 @@ def upgrade():
         'user_tier_history',
         sa.Column('user_id', mysql.MEDIUMINT(display_width=8, unsigned=True),
                   autoincrement=False, nullable=False),
-        sa.Column('tier', mysql.VARCHAR(length=256), nullable=False),
+        sa.Column('tier', mysql.VARCHAR(length=127), nullable=False),
```
Janzert: This file defines base reality for the start of alembic revisions and in general shouldn't be modified unless the definitions actually don't match reality.

Having said that I think all the changes here were a mis-commit as the commit message was "Add mu/sigma/rank fields to game_participant", but only changes to this file were made.

lidavidm: Ah yeah, this is a mis-commit. (I made a local copy of the database, but the MySQL version I used didn't like the length of some of the fields.)

apiserver/apiserver/coordinator/matchmaking.py
lidavidm: The `reset_challenges` function defined above resets any such challenges, though integrating them into the query also makes sense.

Janzert: Yep and I think there is value in keeping the `reset_challenges` button. This just allows them to get picked up even if the full reset isn't done.

lidavidm:
Thanks for the review, I'll update the revision once that update is available. I've also included the revision to add the fields you requested to `game_participant`.

lidavidm:
Alright, I modified the Alembic revision to be current with @Janzert's changes. @harikmenon, are the features here what you expect?

Janzert:
I should have caught this earlier, but the game_participant table already has a rank field for the player's placement in the game. The new field will need to be called something like leaderboard_rank.

lidavidm:
Good catch, I should've checked that more closely.

Janzert:
One other though, might want to hold off on merging this until alembic is fully initialized on production. It will probably make the alembic initialization slightly simpler to have the head version matching what is actually in the database currently. In particular if reality is matching something other than head, make sure the correct version is referenced when issuing the `alembic stamp` command.

harikmenon:
@lidavidm : I will review this with @j-clap today. What do you mean diverts 5% of games to challenges?

Janzert:
5% of the time a worker tries to get a game task it will be used to try and play a challenge game instead of a standard matchmaking game.

harikmenon:
Ha i see, so at-most given our game rate 100 games/hour so 3 (30 games/challenge) challenges will be completed in an hour, which translates to 72 users at max being able to issue atleast one challenge and complete that per day.

Damn that a bit low, can we up this to 10% and I can correspondingly add more instances to our instance group

lidavidm:
Sure, upped to 10%.

apiserver/apiserver/coordinator/matchmaking.py
```diff
@@ -129,9 +135,121 @@ def serve_game_task(conn, has_gpu=False):
...
+                (model.ranked_bots.c.user_id == model.challenge_participants.c.user_id) &
+                (model.challenge_participants.c.challenge_id == challenge["id"]) &
+                (model.challenge_participants.c.user_id == model.ranked_bots_users.c.user_id)
+            )
+        ) &
+        player_filter
+    )
+    bots = conn.execute(bots_query).fetchall()
+
+    # TODO: assumes one-bot-per-player
```
j-clap: Can you handle this TODO? And at least log the occurrence of deviation.

lidavidm: Ok, challenge matchmaking now works with multiple bots per user.

Snaar:
So any reason not to push? Is everyone afraid to hit the merge button? :P

lidavidm:
I'm not sure if there's more wanted - https://github.com/HaliteChallenge/Halite-II/issues/233 at least implies this should support tracking/inviting players who haven't signed up yet, for instance.

Janzert:
Also before merging this it might be best to finish initializing alembic on the production db while the state of repository and the database match.

PR #277, opened 2017-11-11
Refactored julia starter kit

shorty66:
I have looked through all the code and thought the map_string - parsing could need a clean-up.
With this PR i moved all parsing related functions to `parse.jl` and split the parsing in several easier functions. There could be some more documentation, but i think this is a lot easier to understand than before.

I also replaced all String fields in the types to be Integer or Float. This should be good for performance, but is also easier to use in a bot.

I plan to review the types again, as there are still some unused fields, which are not neccessary. But this would likely change the API, breaking the (few) existing bots.

Any comments?


Snaar:
I wouldn't worry about breaking existing bots too much. None of the submitted bots will be affected since they have full copy of the kit in them. In other cases people should know that code/api can change if they go out of their way to get new version of the kit.

airesources/Julia/hlt/constants.jl
```diff
@@ -0,0 +1,31 @@
...
+    WEAPON_RADIUS
+
+# Constants
+const ADDITIONAL_PRODUCTIVITY = 6
+const BASE_PRODUCTIVITY = 6
+const BASE_SHIP_HEALTH = 255
+const DOCKED_SHIP_REGENERATION = 0
+const DOCK_RADIUS = 4.0
+const DOCK_TURNS = 5
+const DRAG = 7.0
```
Snaar: Doesn't exist anymore. Let's remove.

airesources/Julia/hlt/constants.jl
```diff
@@ -0,0 +1,31 @@
...
+const MAX_ACCELERATION = 7.0
+const MAX_SHIP_HEALTH = 255
+const MAX_SPEED = 7.0
+const MAX_TURNS = 300
+const PLANETS_PER_PLAYER = 6
+const PRODUCTION_PER_SHIP = 72
+const RESOURCES_PER_RADIUS = 144
+const SHIPS_PER_PLAYER = 3
+const SHIP_RADIUS = 0.5
+const SPAWN_RADIUS = 2
```
Snaar: 2.0 to make it clear it's a float

airesources/Julia/hlt/constants.jl
```diff
@@ -0,0 +1,31 @@
...
+const BASE_PRODUCTIVITY = 6
+const BASE_SHIP_HEALTH = 255
+const DOCKED_SHIP_REGENERATION = 0
+const DOCK_RADIUS = 4.0
+const DOCK_TURNS = 5
+const DRAG = 7.0
+const EXPLOSION_RADIUS = 10.0
+const EXTRA_PLANETS = 4
+const INFINITE_RESOURCES = true
+const MAX_ACCELERATION = 7.0
```
Snaar: No need for this either.

airesources/Julia/hlt/constants.jl
```diff
@@ -0,0 +1,31 @@
...
+# Constants
+const ADDITIONAL_PRODUCTIVITY = 6
+const BASE_PRODUCTIVITY = 6
+const BASE_SHIP_HEALTH = 255
+const DOCKED_SHIP_REGENERATION = 0
+const DOCK_RADIUS = 4.0
+const DOCK_TURNS = 5
+const DRAG = 7.0
+const EXPLOSION_RADIUS = 10.0
+const EXTRA_PLANETS = 4
```
Snaar: Probably not interesting to bots at all. It's part of map generator.

airesources/Julia/hlt/constants.jl
```diff
@@ -0,0 +1,31 @@
...
+    EXTRA_PLANETS, INFINITE_RESOURCES, MAX_ACCELERATION, MAX_SHIP_HEALTH, MAX_SPEED,
+    MAX_TURNS, PLANETS_PER_PLAYER, PRODUCTION_PER_SHIP, RESOURCES_PER_RADIUS,
+    SHIPS_PER_PLAYER, SHIP_RADIUS, SPAWN_RADIUS, WEAPON_COOLDOWN, WEAPON_DAMAGE, 
+    WEAPON_RADIUS
+
+# Constants
+const ADDITIONAL_PRODUCTIVITY = 6
+const BASE_PRODUCTIVITY = 6
+const BASE_SHIP_HEALTH = 255
+const DOCKED_SHIP_REGENERATION = 0
```
Snaar: Deprecated. Can be removed.

airesources/Julia/hlt/constants.jl
```diff
@@ -0,0 +1,31 @@
+export ADDITIONAL_PRODUCTIVITY, BASE_PRODUCTIVITY, BASE_SHIP_HEALTH, 
+    DOCKED_SHIP_REGENERATION, DOCK_RADIUS, DOCK_TURNS, DRAG, EXPLOSION_RADIUS,
+    EXTRA_PLANETS, INFINITE_RESOURCES, MAX_ACCELERATION, MAX_SHIP_HEALTH, MAX_SPEED,
+    MAX_TURNS, PLANETS_PER_PLAYER, PRODUCTION_PER_SHIP, RESOURCES_PER_RADIUS,
+    SHIPS_PER_PLAYER, SHIP_RADIUS, SPAWN_RADIUS, WEAPON_COOLDOWN, WEAPON_DAMAGE, 
+    WEAPON_RADIUS
+
+# Constants
+const ADDITIONAL_PRODUCTIVITY = 6
```
Snaar: Can be removed and just use BASE_PRODUCTIVITY.

airesources/Julia/hlt/constants.jl
```diff
@@ -0,0 +1,31 @@
...
+const BASE_PRODUCTIVITY = 6
+const BASE_SHIP_HEALTH = 255
+const DOCKED_SHIP_REGENERATION = 0
+const DOCK_RADIUS = 4.0
+const DOCK_TURNS = 5
+const DRAG = 7.0
+const EXPLOSION_RADIUS = 10.0
+const EXTRA_PLANETS = 4
+const INFINITE_RESOURCES = true
```
Snaar: Can be removed, will never be false.

airesources/Julia/hlt/constants.jl
```diff
@@ -0,0 +1,31 @@
...
+const MAX_ACCELERATION = 7.0
+const MAX_SHIP_HEALTH = 255
+const MAX_SPEED = 7.0
+const MAX_TURNS = 300
+const PLANETS_PER_PLAYER = 6
+const PRODUCTION_PER_SHIP = 72
+const RESOURCES_PER_RADIUS = 144
```
Snaar: Probably not interesting as well.

airesources/Julia/hlt/constants.jl
```diff
@@ -0,0 +1,31 @@
...
+const DOCK_TURNS = 5
+const DRAG = 7.0
+const EXPLOSION_RADIUS = 10.0
+const EXTRA_PLANETS = 4
+const INFINITE_RESOURCES = true
+const MAX_ACCELERATION = 7.0
+const MAX_SHIP_HEALTH = 255
+const MAX_SPEED = 7.0
+const MAX_TURNS = 300
+const PLANETS_PER_PLAYER = 6
```
Snaar: Part of mapgen constants, probably not interesting to bot.

airesources/Julia/hlt/constants.jl
```diff
@@ -0,0 +1,31 @@
...
+const DOCKED_SHIP_REGENERATION = 0
+const DOCK_RADIUS = 4.0
+const DOCK_TURNS = 5
+const DRAG = 7.0
+const EXPLOSION_RADIUS = 10.0
+const EXTRA_PLANETS = 4
+const INFINITE_RESOURCES = true
+const MAX_ACCELERATION = 7.0
+const MAX_SHIP_HEALTH = 255
+const MAX_SPEED = 7.0
```
Snaar: This is technically an integer, since speeds are sent as integers to the game engine. I would prefer for it to be "7".

shorty66:
You are right, breaking bots shouldn't be a problem. I'll work a bit more on this, then.

Snaar:
This looks pretty good already, so I'll merge it in.

PR #278, opened 2017-11-11
Credit players for collision damage (fixes #275)

### LINKED ISSUE #275: Count collision damage
https://forums.halite.io/t/collision-damage-not-counted-for/490/2

lidavidm:


Janzert:
Currently this makes it so that when p1 crashes his ship into a ship of p2, p1 is credited with damage equal to the remaining health of p2's ship and p2 with p1's ship. I think this uncontroversial and the primary purpose for this change.

Now similarly when p1 crashes a ship into p2's planet, p1 is credited with the amount of damage caused to the planet and p2 is credited with the ship's health at the time. Here though I don't think p2 should be getting credit for damage given. First because the player can't intentionally act to crash their planet into a ship. Second it makes sense to me that although you may 'own' the planet it is still a contested resource possibly taken over by another player at any point and fundamentally different from owning a ship. As such no damage caused by it should go to either player, not the damage from ships crashing into it or the damage caused if it explodes.

fohristiwhirl:
I agree with Janzert.

lidavidm:
Ok, sounds good. I'll change it to only account for damage dealt to & from ships.

environment/core/Halite.cpp
```diff
@@ -719,10 +721,53 @@ auto Halite::process_events() -> void {
             target_count[src] = num_prev_targets + 1;
         };
 
+        auto credit_damage = [&](hlt::EntityId source,
+                                 hlt::EntityId target,
+                                 int damage) -> void {
+            hlt::PlayerId dealer = 0;
+            hlt::PlayerId receiver = 0;
+            switch (source.type) {
+            case hlt::EntityType::PlanetEntity: {
```
Janzert: For the way I think damage should be accounted (each user's ship to anything), just remove this case statement should work.

lidavidm:
Updated.

PR #281, opened 2017-11-12
Bring repo alembic revisions and sql schema in line with reality

Janzert:
* Add the new leagues table in the database.
* Drops the change to game id type for now (since it hasn't been implemented in production yet).
   I'll re-add this in another PR that can be left unmerged until it's ready to be used in production.
* Changes schema.sql to use mysqldump output and adds a script to regenerate it.
   There are a couple of reasons I decided to switch to this format.
     * This makes it an easy to reference snapshot of the current database state.
     * Users that are unfamiliar with alembic can modify it directly and then someone with alembic experience can create necessary alembic version.
     * It can be easily used to compare with an actual database


Janzert:
I forgot to mention that I'd be happy to help walk through whoever it is that will do the final initialization of alembic on production. Although at this point I think it boils down to:

* Get alembic configured in a location that has database access
    configuration can be done through `apiserver/apiserver/config.py` or `apiserver/alembic.ini`
* As with anything that modifies the database, make a database backup before starting
* Use the `apiserver/sql/generate_schema.sh` script to compare the production schema with the expected schema in `apiserver/sql/schema.sql`.
    The only expected differences are `mysqldump` header and dump time comments and autoincrement values.
* run `alembic stamp head` to initialize the alembic version table.

PR #282, opened 2017-11-12
rewrite make.bat for halite and client

Snaar:


Snaar:
Made into PR to raise awareness that this is going in, in case Windows build breaks or there is feedback later.

environment/make.bat
```diff
@@ -1,6 +1,40 @@
...
+if not defined vcvarsall_location (
+    echo Failed to find vcvarsall.bat in any of the known places. You have two options:
+    echo 1^) Preferred: run vcvarsall.bat yourself before running this script. Check out https://docs.microsoft.com/en-us/cpp/build/building-on-the-command-line for more information.
+    echo 2^) Find where vcvarsall.bat file is on your system and add it to the list of locations in this batch file.
+    pause
+    exit /b 1
+)
+
+reg query "HKLM\SYSTEM\CurrentControlSet\Control\Session Manager\Environment" /v PROCESSOR_ARCHITECTURE | find /i "x86" > nul
+if !ERRORLEVEL! == 0 (
```
Janzert: I'm guessing that the systems used to build the 32bit binaries currently are actually 64bit systems. So there probably needs to be a way to manually override this.

Snaar: You'd run vcvarsall.bat yourself before calling this script, which is how MS intended you use this.

Snaar: Also we don't actually build any 32 bit binaries anywhere ourselves. Some people asked for being able to build on 32 bit windows, since that's the only thing they had. On those machines it should "just work".

Janzert: Oops, sorry I thought I had seen a 32bit download for windows before. Must have just been the .bat file.

PR #288, opened 2017-11-13
Explicitly notify user if compiled bot too large

### LINKED ISSUE #287: Bots failing to compile
https://halite.io/user/?user_id=1224

https://github.com/ruyimarone

lidavidm:
Fixes #287

### DIFF
--- a/apiserver/apiserver/config.py
+++ b/apiserver/apiserver/config.py
@@ -16,6 +16,7 @@ COMPILATION_STUCK_THRESHOLD = 30
 # Flask settings
 # Max size of an upload, in bytes
 MAX_BOT_UPLOAD_SIZE = 20 * 1024 * 1024
+# Needs to match corresponding value in apiserver/config.py
 MAX_COMPILED_BOT_UPLOAD_SIZE = 100 * 1024 * 1024
 # Secret key for Flask session cookies
 FLASK_SECRET_KEY = ""
--- a/apiserver/apiserver/config.py
+++ b/apiserver/apiserver/config.py
@@ -16,7 +16,7 @@ COMPILATION_STUCK_THRESHOLD = 30
 # Flask settings
 # Max size of an upload, in bytes
 MAX_BOT_UPLOAD_SIZE = 20 * 1024 * 1024
-# Needs to match corresponding value in apiserver/config.py
+# Needs to match corresponding value in worker configuration
 MAX_COMPILED_BOT_UPLOAD_SIZE = 100 * 1024 * 1024
 # Secret key for Flask session cookies
 FLASK_SECRET_KEY = ""
--- a/apiserver/worker/backend.py
+++ b/apiserver/worker/backend.py
@@ -7,7 +7,7 @@ from time import gmtime, strftime, sleep
 
 
 # Needs to match corresponding value in apiserver/config.py
-# Default value, 100 MiB
+# This is the default value, 100 MiB
 MAX_BOT_UPLOAD_SIZE = 100 * 1024 * 1024
 # Maximum wait time in between compiled bot archive upload attempts,
 # in seconds
@@ -19,8 +19,9 @@ with open("config.json") as configfile:
     MANAGER_URL = config["MANAGER_URL"]
     SECRET_FOLDER = config["SECRET_FOLDER"]
     CAPABILITIES = config.get("CAPABILITIES", [])
-    MAX_BOT_UPLOAD_SIZE = config.get("MAX_BOT_UPLOAD_SIZE",
-                                     MAX_BOT_UPLOAD_SIZE)
+    provided_size = config.get("MAX_BOT_UPLOAD_SIZE", MAX_BOT_UPLOAD_SIZE)
+    if provided_size:
+        MAX_BOT_UPLOAD_SIZE = provided_size
 
 
 def getTask():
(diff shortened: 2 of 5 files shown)


apiserver/worker/worker.py
```diff
@@ -62,6 +62,15 @@
...
 
 
+UPLOAD_ERROR_MESSAGE = """
+We had some trouble uploading your bot; likely the final compiled bot
+was too large (> 100 MiB). If you cannot figure out why this happened,
+please email us at halite@halite.io. We can help.
+
+For our reference, here is the trace of the error:
+"""
+
```
Janzert: Given that the size is now explicitly checked before upload, I think I'd leave out the part about "likely the final compiled bot was too large".

apiserver/worker/backend.py
```diff
@@ -6,6 +6,9 @@
 from time import gmtime, strftime
 
 
+MAX_BOT_UPLOAD_SIZE = 100 * 1024 * 1024
```
Janzert: Might want to leave a note here and in `apiserver/worker/config.py` near `MAX_COMPILED_BOT_UPLOAD_SIZE` that the values should match.

apiserver/worker/backend.py
```diff
@@ -95,6 +101,8 @@ def storeBotRemotely(user_id, bot_id, zip_file_path):
                           },
                           files={"bot.zip": zip_contents})
         print("Posting compiled bot archive %s\n" % r.text)
+        # We don't check the status code since we instaed check the
+        # bot hash.
```
Janzert: There are many error codes that are unlikely to resolve on retry (almost by definition pretty much all the 4xx codes) and we could bail out without trying another 99 times. Also any response error that is likely temporary probably deserves a backoff before retrying.

But leaving it for now is fine, the other changes definitely make it much better than what happens now.

lidavidm:
Thanks for the feedback. I changed it to back off uploads, and reduced the number of tries to 10 consequently.

apiserver/apiserver/config.py
```diff
@@ -16,6 +16,7 @@
 # Flask settings
 # Max size of an upload, in bytes
 MAX_BOT_UPLOAD_SIZE = 20 * 1024 * 1024
+# Needs to match corresponding value in apiserver/config.py
```
j-clap: Why does this need to match the value in itself? IRL recursion :P

apiserver/worker/backend.py
```diff
@@ -3,7 +3,11 @@
 from hashlib import md5
 import json
 import os
-from time import gmtime, strftime
+from time import gmtime, strftime, sleep
+
+
+# Needs to match corresponding value in apiserver/config.py
+MAX_BOT_UPLOAD_SIZE = 100 * 1024 * 1024
```
j-clap: Rather than define it twice can't wejust pick it from there/another central location for both?

apiserver/worker/backend.py
```diff
@@ -84,22 +88,33 @@ def storeBotLocally(user_id, bot_id, storage_dir, is_compile=False):
...
+        if r.status_code >= 400 and r.status_code <= 499:
+            print("Got a 4xx status code")
+            r.raise_for_status()
 
         # Try again if local and remote hashes differ
         if local_hash != getBotHash(user_id, bot_id):
             print("Hashes do not match! Redoing file upload...\n")
             iterations += 1
+            sleep(backoff)
+            if backoff < 32:
```
j-clap: Magic variable.

j-clap:
Added small comments. I'm up for approving when they're addressed :)

apiserver/worker/backend.py
lidavidm: Yeah, the issue is that the worker doesn't assume it has access to the server codebase (though this happens to be true currently). Perhaps we can make it part of the `config.json`, or have a server endpoint that returns the value?

j-clap: Config.json would be fine :)

lidavidm:
Updated, thanks for the review!

apiserver/worker/backend.py
```diff
@@ -3,14 +3,24 @@
 from hashlib import md5
 import json
 import os
-from time import gmtime, strftime
+from time import gmtime, strftime, sleep
+
+
+# Needs to match corresponding value in apiserver/config.py
+# Default value, 100 MiB
+MAX_BOT_UPLOAD_SIZE = 100 * 1024 * 1024
```
Janzert: At first I thought this was now redundant, but I see it's used as a default when the config doesn't define a value. The comment should probably be updated to reflect that.

apiserver/worker/backend.py
```diff
@@ -3,14 +3,24 @@
...
+MAX_UPLOAD_BACKOFF = 32
 
 
 with open("config.json") as configfile:
     config = json.load(configfile)
     MANAGER_URL = config["MANAGER_URL"]
     SECRET_FOLDER = config["SECRET_FOLDER"]
     CAPABILITIES = config.get("CAPABILITIES", [])
+    MAX_BOT_UPLOAD_SIZE = config.get("MAX_BOT_UPLOAD_SIZE",
+                                     MAX_BOT_UPLOAD_SIZE)
```
Janzert: This should probably keep the default value if the config value is None.

Issue #71, opened 2017-10-25
Replay files download through halite clients doesn’t function correctly


harikmenon:
https://forums.halite.io/t/replay-files-download-through-halite-clients-doesnt-function-correctly/203

harikmenon:
Closing this issue https://forums.halite.io/t/how-do-you-open-hlt-files/243/5

PR #289, opened 2017-11-13
`hlt replay` saving as valid json

kjschiroo:
I believe this is the underlying issue for #71. The `hlt` client doesn't save valid json.

Previously binary strings were being cast to a string using `str`. This generates the string
```
b'{ real_json}'
```
The binary string needs to be decoded rather than cast. This results in the value
```
{ real_json }
```
The [work around](https://forums.halite.io/t/how-do-you-open-hlt-files/243/2?u=kjschiroo) the community has been using is to remove the first 2 and last byte, then parse it.

### DIFF
--- a/tools/hlt_client/hlt_client/download_game.py
+++ b/tools/hlt_client/hlt_client/download_game.py
@@ -62,7 +62,7 @@ class GameDownloader:
         :return: the file unzipped if possible
         """
         try:
-            return str(zstd.loads(game_binary))
+            return zstd.loads(game_binary).decode()
         except Exception:
             raise ValueError("Could not unzip file at: {}!".format(game_id))
 
@@ -199,4 +199,3 @@ def download(mode, destination, date, all_bots, default_user_id, user_id, limit)
             raise ValueError("Cannot run default mode without authenticating .Please run `client.py --auth` first.")
         UserGameDownloader(destination, default_user_id if not user_id else user_id, limit).get_objects()
     print('Finished writing files to desired location')
-


j-clap:
(Also, updated pip version)

LecJackS:
Couldn't make _zstd.decompress()_ work ([maybe it was compressed with an older version and now it needs to be decompressed with _decompress_old()_?](https://github.com/sergey-dryabzhinsky/python-zstd#warning)) or the removing solution right away.

I'll add another alternative here, basically the same as [the one from csheldrik](https://forums.halite.io/t/replay-files-download-through-halite-clients-doesnt-function-correctly/203/5?u=lecjacks) , but with read().decode() instead of readlines() so it works on my system:

```python
def fetch_data_dir(directory, limit):
    """
    Loads up to limit games into Python dictionaries from uncompressed replay files.
    """
    replay_files = sorted([f for f in os.listdir(directory) if
                           os.path.isfile(os.path.join(directory, f)) and f.startswith("halite-2-gold-replays_replay")])

    if len(replay_files) == 0:
        raise Exception("Didn't find any game replays. Please call make games.")

    print("Found {} games.".format(len(replay_files)))
    print("Trying to load up to {} games ...".format(limit))

    loaded_games = 0

    all_data = []
    for r in replay_files:
        full_path = os.path.join(directory, r)
        with open(full_path, "rb") as rpfile:
            print(loaded_games+1, full_path)
            game_data = rpfile.read().decode('utf-8')
            if game_data[0:2] == "b'":
                #if the first two decoded characters are b'
                game_data = game_data[2:-1]
            game_json_data = json.loads(game_data)
            all_data.append(game_json_data)
            loaded_games = loaded_games + 1

        if (loaded_games >= limit) or (loaded_games >= len(replay_files)):
            """Loads just the amount of game files stablished at console
               but never more than the amount of game files available"""
            break

    print("{} games loaded.".format(loaded_games))

    return all_data
```

Issue #302, opened 2017-11-15
Compile with CMakeLists.txt if present in bot submission


Snaar:
Inspired by #249. Looks like while we did add CMakeLists.txt to the cpp kit, the backend was not updated. We should keep existing build logic in case CMakeLists.txt is not found, but we should add one that uses cmake/make otherwise.

Snaar:
This target/language/whatever can be called CMake as opposed to pure C/C++.

Snaar:
I'll give it a try.

Snaar:
@a-price, once @harikmenon updates the worker image the "boost in bot root directory" should work.

PR #309, opened 2017-11-16
adding cmake support to worker compiler (#302)

Snaar:


apiserver/worker/compiler.py
```diff
@@ -43,10 +43,10 @@ def __exit__(self, type, value, traceback):
 def safeglob(pattern):
     safepaths = []
     for root, dirs, files in os.walk("."):
-        print("Walking: " + root + " " + ", ".join(dirs) + " " + ", ".join(files))
+        # print("Walking: " + root + " " + ", ".join(dirs) + " " + ", ".join(files))
```
Snaar: this was just too verbose and didn't seem to have good reason to stay verbose

apiserver/worker/compiler.py
```diff
@@ -43,10 +43,10 @@ def __exit__(self, type, value, traceback):
...
         files = fnmatch.filter(files, pattern)
         for fname in files:
-            print(os.path.splitext(fname)[0])
+            # print(fname + " -> " + os.path.splitext(fname)[0])
```
Snaar: ditto re verboseness

Janzert:
It's unfortunate that any bot using it will have the language set to `CMake` instead of their actual language. But I don't see a good way to fix it within the current framework. This whole file really needs a good redesign, but not at this point in the contest.

Janzert:
It looks like the `cmake` package is already being installed on the workers and `make` is almost certainly getting pulled in as a dependency for something. It might be a good idea to confirm that both are installed on the workers though.

Snaar:
Yeah, setup_worker_image.sh suggests cmake is there, but I'm actually not sure how to check empirically. I guess I should figure that out or someone else can :P

Snaar:
All right, it's going to report itself as C++ now.

apiserver/worker/compiler.py
```diff
@@ -278,6 +278,10 @@ def compile(self, bot_dir, globs, errors, timelimit):
         ["gnatbind"],
         ["gnatlink", "-o", BOT],
     ],
+    "CMake": [
+        ["cmake", "."],
+        ["make", "-j2"],
```
Snaar: How many cores do workers get for compilation? If it's 1 I'll change this.

apiserver/worker/compiler.py
```diff
@@ -675,6 +675,10 @@ def compile_anything(bot_dir, installTimeLimit=600, timelimit=600, max_error_len
                 print("error")
                 print(e)
 
+            # let's assume that anyone using cmake is using it for C++
+            if name == "CMake":
+                name = "C++"
+
```
Janzert: Instead of a second special casing for cmake could it be combined into the extra cmake conditional at line 610, just change the lang.name there?

apiserver/worker/compiler.py
Janzert: It should have access to all 4 cores during compilation

apiserver/worker/compiler.py
Snaar: Unfortunately no, because compilation happens right between these two places.

Snaar: (And you need CMake for compilation.)

Snaar: Though hopefully I'm misunderstanding what you are proposing and it _can_ be made nicer.

Snaar:
Yeah, let's see if @lidavidm or @harikmenon think those print statements are useful when debugging compilation issue. If not I'll remove them.

Snaar:
(Though they don't really do anything but print the contents of the pack couple times.)

lidavidm:
CMake 3.7 is installed.

PR #310, opened 2017-11-16
Remove mentions of run.sh.

Janzert:
As `run.sh` isn't fully implemented (and never was from what I can find) this removes mention of it from the documentation so that people aren't led astray trying to make it work.


Janzert:
As a side note, since somewhere in the neighborhood of 2 or 3 people have mentioned attempting to use `run.sh` I'm not sure it's worth putting much effort into making it work. A simple workaround for those that do need it is a `MyBot.py` that launches the actual bot code. Something like the following should work fine:

    BOT_CMD = "./MyBot"
    import subprocess
    subprocess.run(BOT_CMD)

PR #313, opened 2017-11-17
Properly kill child processes

JonathanRaiman:
Process kill call was negating the pid, and not properly killing children.

### DIFF
--- a/environment/networking/Networking.cpp
+++ b/environment/networking/Networking.cpp
@@ -695,7 +695,7 @@ void Networking::kill_player(hlt::PlayerId player_tag) {
         } else break;
     }
 
-    kill(-processes[player_tag], SIGKILL);
+    kill(processes[player_tag], SIGKILL);
 
     processes[player_tag] = -1;
     connections[player_tag].read = -1;


environment/networking/Networking.cpp
```diff
@@ -695,7 +695,7 @@ void Networking::kill_player(hlt::PlayerId player_tag) {
         } else break;
     }
 
-    kill(-processes[player_tag], SIGKILL);
+    kill(processes[player_tag], SIGKILL);
```
Snaar: Hmm. This is supposed to kill the process and all the children. I imagine some OSes don't handle this? How about this, do both kills - one with negative to kill tree where it's supported, and then kill with non-negative if kill fails?

Snaar: Also, what OS did you have this issue on? Just for reference.

Snaar: I.e. if kill(-pid) returns < 0, then try kill(pid).

JonathanRaiman: Ah good call -- doing this on MacOS / ubuntu leaves many dangling processes [enough to kill a machine after some time]. Will check on the return code & post back in a minute or so

JonathanRaiman: I'm getting return code 0 for negative pid or positive pid:
```cpp
auto rcode = kill(processes[player_tag], SIGKILL);
std::cout << "killing pos rcode was " << rcode << std::endl;
rcode = kill(-processes[player_tag], SIGKILL);
std::cout << "killing neg rcode was " << rcode << std::endl;
```

Not sure what the root cause is, or if this fixes it at all.
if this isn't the right way, any ideas as to why some processes would not be exiting properly?

Snaar: Does killing with positive code help you at all empirically?

JonathanRaiman: It did, but I'm having a hard time reproducing the issue now.

Snaar:
I'm going to close for now until someone comes in with more data.

PR #314, opened 2017-11-17
adds gradle task to build submittable zip

dorzey:
By running the gradle submission task the submittable zip file is outputted to /build/distributions


Snaar:
:+1:

PR #317, opened 2017-11-18
dartlang starter kit

jerold:
Allows the player to write their bot in Dart.

Installing Dart: https://www.dartlang.org/install/linux


airesources/Dart/lib/src/logger.dart
```diff
@@ -0,0 +1,13 @@
...
+
+
+class Logger {
+  String _fileName;
+
+  Logger._();
+
+  void call(String log) {
+    if (_fileName == null) return;
+    new File(_fileName).writeAsStringSync("$log\n", mode: FileMode.WRITE_ONLY_APPEND);
```
Snaar: Let's use WRITE_ONLY mode. Not overwriting log by default would be confusing. All other kits overwrite.

airesources/Dart/lib/src/position.dart
```diff
@@ -0,0 +1,41 @@
...
+    final double x = target.x + radius * cos(angleRad);
+    final double y = target.y + radius * sin(angleRad);
+    return new Position(x: x, y: y);
+  }
+}
+
+int toDegrees(double radians) => (radians * 180 / PI).round();
+
+int angleRadToDegClipped(double angleRad) {
+  return ((toDegrees(angleRad) % 360) + 360) % 360;
```
Snaar: Dart's mod operator always produces positive numbers. So `return toDegrees(angleRad) % 360;` should be sufficient.

Snaar:
Ok this looks good, just say when you are done with updates you want to do for this PR.

jerold:
Should be good to go. Thanks @Snaar for the review!

jerold:
All done with updates. 😀

PR #319, opened 2017-11-20
Perform internal computations in long double

lidavidm:
This is an attempted fix for #135. However, it makes the values in the replay quite misleading; under the case that @shummie submitted, while the engine now correctly computes that ships 9 and 11 are far enough apart to not collide, the values in the JSON suggest they should have collided, even with the tweak suggested in https://github.com/nlohmann/json/issues/360.

### DIFF
--- a/environment/core/Entity.cpp
+++ b/environment/core/Entity.cpp
@@ -2,11 +2,11 @@
 #include "hlt.hpp"
 
 namespace hlt {
-    auto Location::distance(const Location &other) const -> double {
+    auto Location::distance(const Location &other) const -> long double {
         return sqrt(distance2(other));
     }
 
-    auto Location::distance2(const Location &other) const -> double {
+    auto Location::distance2(const Location &other) const -> long double {
         return std::pow(other.pos_x - pos_x, 2) +
             std::pow(other.pos_y - pos_y, 2);
     }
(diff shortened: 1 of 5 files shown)


j-clap:
I must say I highly disagree with this change, considering this will now become a rabbit hole of how far we do consider the mantissa (what's next? Quadruple precision ieee-754? ). I would rather us set a predictable value as a standard and not budge from there. Also note that this can be unpredictable across arches, since cpp only guarantees long double to be as large as double (technically double is only as large as first by the standard as well, but I've found it more consistant); not that it matters to us since wee only on one arch, but good to note.

lidavidm:
Another possible fix is to explicitly check collisions again after updating positions (since the issue here is that the ship moves slightly further than the magnitude of the velocity suggests), but that feels janky to me as well.

j-clap:
That being said, though I disagree with the idea, I will approve since we're already here and the code is already changed, and no detrimental effect should arise.

j-clap:
Yes, they both have issues. I'm not sure this will solve the issue entirely though. But no harm in trying :)

lidavidm:
Yeah, this will definitely not fully solve the issue either, just make it (hopefully) less likely (though I don't think it was all that likely to begin with).

Snaar:
Maybe next time fixed point math :P

j-clap:
We used to have it, then people complained :(

lidavidm:
It was more like perform double computations -> round positions to integers (and IIRC velocities (when we had inertia still) were still doubles) so that was a weird mix.

PR #320, opened 2017-11-22
Update the Rust starter kit

Kerollmops:
Sometimes I follow the api-guideline https://rust-lang-nursery.github.io/api-guidelines/naming.html#c-getter


airesources/Rust/src/hlt/game.rs
```diff
@@ -41,27 +40,29 @@ impl Game {
 
         println!("{}", name);
```
Snaar: Would you mind attempting to address https://github.com/HaliteChallenge/Halite-II/issues/151 at the same time?

Kerollmops: I will fix that yes, thanks for the remark.

airesources/Rust/src/hlt/game.rs
```diff
@@ -41,27 +40,29 @@ impl Game {
...
 
-        let game = Game {
+        // ignore first line
+        Game::read_line();
```
Kerollmops: Just to know, why did we need to skip the first map update, here ?

Snaar: I think original author didn't understand why we needed it. This original update is what you actually need to implement #151 correctly.

Kerollmops:
The real way to create a game should use [a builder](https://rust-lang-nursery.github.io/api-guidelines/type-safety.html#c-builder) and don't let the user own a complete `Game` type once he hasn't send it's username to the game.

This will avoid the user to send command before sending it's own name.

But this way works fine.

PR #323, opened 2017-11-23
Initial work on live game feed.

Janzert:
It's purposefully not linked to from anywhere and could use a few more features (such as a pause button). But I think it's in a state that would be useful to go live. Even if only to get some real world testing from a few more people (and browsers types :) ).

Locally tested with firefox and chrome, both of them on desktop and mobile.

Screenshot:
image

### DIFF
--- a/website/javascript/templates/GameFeed.vue
+++ b/website/javascript/templates/GameFeed.vue
@@ -1,0 +1,248 @@
+<template>
+  <div class="game-feed">
+    <div v-if="display_games.length > 0" class="table-container">
+      <table class="table table-leader">
+        <thead>
+          <tr>
+            <th class="watch">Watch</th>
+            <th class="result">Result</th>
+            <th class="destroyed text-center hidden-xs">Destroyed Ships</th>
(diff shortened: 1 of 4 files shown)


website/javascript/templates/GameFeed.vue
```diff
@@ -0,0 +1,248 @@
...
+        position: absolute;
+        bottom: 0;
+    }
+    .game-participant {
+        img {
+            height: 20px;
+            width: 20px;
+        }
+        .username {
+            font-size: 1rem;
```
lidavidm: At least for me, this makes the font size nearly unreadably tiny, and doesn't seem necessary. For mobile, it doesn't fit all that well anyways and is even tinier - maybe we could save some space by shortening the date/time format, if that's the concern.

Janzert: This wasn't even so much about space as wanting to de-emphasize the usernames, the default 1.6rem feels really heavy. But yeah I probably overdid it, I'll bump it back up some.

Issue #321, opened 2017-11-22
Installation leads to `multiarray numpy extension module failed`


halitesubmission:
My halite solution is currently having trouble running (crashes on start).
I added a setup.py file that compiles some C++ and cython code, but do not get a compilation error.
My bot is then run using "MyBot.py" (e.g. autodetected as Python), but apparently fails on run by throwing `ImportError: Importing the multiarray numpy extension module failed.`

Has anyone run into these issues/found a reason why compiling can corrupt the numpy installation?

halitesubmission:
This issue is in fact due to running Python 3.5 instead of 3.6 [possibly vice-versa, not sure].
In any case, if the *right* Python is run (3.6), then this issue goes away.

More pressing:
if install.sh fails during compilation this is not flagged, and no stderr from that process is ever given back to the user. Could someone add this to one of the capabilities [or indicate how to get python install.sh files to be properly detected as working/failing].

My current solution is to:
1. redirecting std-err to a file
2. Use errors as bot name
3. see error logs's bot name to read errors

obviously 2 & 3 are not ideal. Any suggestions?

lidavidm:
Yes, `python3` points to `python3.5` by default, unfortunately (since we're using an older version of Ubuntu). I'll look into getting the output of install.sh included in the log that is sent to you.

halitesubmission:
Thanks. Change seems to add errors from install.sh to the rest of the logs

PR #324, opened 2017-11-23
Include result of install.sh in error logs

lidavidm:
Makes #321 easier to debug.

### DIFF
--- a/apiserver/worker/compiler.py
+++ b/apiserver/worker/compiler.py
@@ -772,7 +772,7 @@ def compile_anything(bot_dir, installTimeLimit=600, timelimit=600, max_error_len
     detected_language, language_errors = detect_language(bot_dir)
 
     print("detected language " + str(detected_language))
-    if not detected_language or install_errors:
+    if not detected_language:
         return "Unknown", truncate_errors(install_stdout, install_errors,
                                           language_errors, [], max_error_len)
 
--- a/apiserver/worker/compiler.py
+++ b/apiserver/worker/compiler.py
@@ -680,14 +680,30 @@ def get_run_lang(submission_dir):
                     if line[0] == '#':
                         return line[1:-1]
 
+
 def compile_anything(bot_dir, installTimeLimit=600, timelimit=600, max_error_len = 3072):
+    errors = []
     if os.path.exists(os.path.join(bot_dir, "install.sh")):
-        _, errors = _run_cmd("chmod +x install.sh; ./install.sh", bot_dir, installTimeLimit)
-    detected_language, errors = detect_language(bot_dir)
+        install_stdout, install_errors = _run_cmd(
+            "chmod +x install.sh; ./install.sh", bot_dir, installTimeLimit)
+        if install_errors:
+            errors.append("Possible errors running install.sh. stderr output as follows:\n")
+            errors.extend(install_errors)
+            errors.append("\nEnd of install.sh stderr. stdout as follows:\n")
+            errors.extend(install_stdout)
+            errors.append("\nEnd of install.sh stdout.\n")
+
+    detected_language, language_errors = detect_language(bot_dir)
+    if language_errors:
+        errors.extend(language_errors)
+
     print("detected language " + str(detected_language))
     if detected_language:
         print("compiling")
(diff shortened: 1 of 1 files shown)


Janzert:
errors is chopped up right after this section in order to keep the size of it under 3kb, with 1kb coming from the start of errors and 2kb from the end. My guess is that the most useful part of the output is likely to be the end of stderr, unfortunately there is probably a decent chance with this system of that getting removed. The separation messages, "end of install.sh stderr, stdout follows", etc., are also likely to disappear.

The best fix would probably be to keep them all separate until the trimming is done, trim them each separately (maybe with more weight to stderr output over stdout) and then combining with the separation messages. Also the limit should be fine to increase to say 10kb without causing any issue and give a better chance of keeping the important bits.

lidavidm:
@Janzert good idea, I've implemented your suggestion.

apiserver/worker/compiler.py
```diff
@@ -680,63 +680,139 @@ def get_run_lang(submission_dir):
                     if line[0] == '#':
                         return line[1:-1]
 
-def compile_anything(bot_dir, installTimeLimit=600, timelimit=600, max_error_len = 3072):
+
+INSTALL_ERROR_START = "Possible errors running install.sh. stdout output as follows:"
+INSTALL_ERROR_MID = "End of install.sh stdout. stderr as follows:"
+INSTALL_ERROR_END = "End of install.sh stdout."
```
Janzert: INSTALL_ERROR_END isn't put at the end of the install.sh stdout so instead should probably be something like "End of install.sh output."

apiserver/worker/compiler.py
```diff
@@ -680,63 +680,139 @@ def get_run_lang(submission_dir):
...
-                    end_errors.append(errors[cur_error])
-                    cur_error -= 1
-                    length += len(errors[cur_error])
-                end_errors.reverse()
-                errors = first_errors + end_errors
-
-            return detected_language.name, errors
-    else:
-        return "Unknown", errors
+    if not detected_language or install_errors:
```
Janzert: I think bailing out here if anything was written to stderr in install.sh is too eager, some programs simply log informational messages to stderr. Currently the compilation is still attempted and only fails if the actual compilation step fails.

apiserver/worker/compiler.py
```diff
@@ -680,63 +680,139 @@ def get_run_lang(submission_dir):
...
+    install_errors = install_errors or []
+    language_detection_errors = language_detection_errors or []
+    compile_errors = compile_errors or []
+
+    all_errors = language_detection_errors + compile_errors
+    result = []
+    if install_errors:
+        all_errors = install_stdout + install_errors
+
+    if sum(len(line) for line in all_errors) <= max_error_len:
```
Janzert: This isn't including the lengths of langauage_detection and compile errors. I think language detection should always be short but compile errors can be quite long.

apiserver/worker/compiler.py
```diff
@@ -680,63 +680,139 @@ def get_run_lang(submission_dir):
...
+        if install_errors:
```
Janzert: Even if there was nothing on stderr from install.sh, the stdout should probably still be included.

apiserver/worker/compiler.py
```diff
@@ -680,63 +680,139 @@ def get_run_lang(submission_dir):
...
+                end_errors.append(source[current])
+                length += len(source[current])
+                current -= 1
+            result.extend(reversed(end_errors))
+
+        return length, result
+
+
+    remaining_length = max_error_len
+    if install_errors:
```
Janzert: Same as above, probably want install.sh stdout even when no stderr output.

apiserver/worker/compiler.py
Janzert: Maybe checking the exit code from install.sh would work, but I'd be a little worried about even doing that given that we haven't up till now.

apiserver/worker/compiler.py
Janzert: Oops, misread the code here. It is including language detection and compile errors in all_errors.

lidavidm:
Thanks for the review, I've addressed those comments.

Issue #265, opened 2017-11-10
Haskell : stack support


wizzup:
Following discussion on [forum](http://forums.halite.io/t/starter-kit-updates/380/8?u=wizzup)

There is no details description for Haskell. For example:

* What is the compiler version? (`ghc --version`)
* What are libraries available to use? (`ghc-pkg list`)
* What about package management helper? (cabal / stack)

I would love to request for [stack](https://docs.haskellstack.org/en/stable/README/) installation. 

If the official installation method does not applicable, you may adapt instructions from my Halite1 [PR](https://github.com/HaliteChallenge/Halite/pull/282/commits/908e3fbc23aa81aed873570aba0ff53a51a0bdf5)

Additionally, all language should include ecosystem detail in startup bot's README.md

lidavidm:
GHC:
```
$ ghc --version
The Glorious Glasgow Haskell Compilation System, version 8.0.2
$ ghc-pkg list
/usr/lib/ghc/package.conf.d
    Cabal-1.24.2.0
    array-0.5.1.1
    base-4.9.1.0
    binary-0.8.3.0
    bytestring-0.10.8.1
    containers-0.5.7.1
    deepseq-1.4.2.0
    directory-1.3.0.0
    filepath-1.4.1.1
    ghc-8.0.2
    ghc-boot-8.0.2
    ghc-boot-th-8.0.2
    ghc-prim-0.5.0.0
    ghci-8.0.2
    haskeline-0.7.3.0
    hoopl-3.10.2.1
    hpc-0.6.0.3
    integer-gmp-1.0.0.1
    pretty-1.1.3.3
    process-1.4.3.0
    rts-1.0
    template-haskell-2.11.1.0
    terminfo-0.4.0.2
    time-1.6.0.1
    transformers-0.5.2.0
    unix-2.7.2.1
    xhtml-3000.2.1
```

Neither Cabal nor Stack are available, but we can install those.

lidavidm:
I will install these versions, and once we rebuild the worker images and recreate the workers they should be available:

```
$ stack --version
Version 1.1.2 x86_64 hpack-0.14.1
$ cabal --version
cabal-install version 1.24.0.1
compiled using version 1.24.1.0 of the Cabal library
```

Let us know if there is other stuff that would be useful for Haskell. (This is what is available out-of-the-box on Ubuntu 17.04.)

wizzup:
[...]

PR #325, opened 2017-11-24
Add Stack support for Haskell

lidavidm:


apiserver/worker/compiler.py
```diff
@@ -489,6 +493,14 @@ def compile(self, bot_dir, globs, errors, timelimit):
         [BOT],
         [([""], ExternalCompiler(comp_args["Haskell"][0]))]
     ),
+    Language("Haskell/Stack", BOT, "stack.yaml",
```
Janzert: Would a bot using still be likely to include a `MyBot.hs` file in the root directory of the zip? If it does you'll need to add a language disambiguation section in `dectect_language` like Snaar did for cmake based builds.

lidavidm: Based on what @wizzup provided in https://github.com/HaliteChallenge/Halite-II/issues/265#issuecomment-346775263, it seems not.

Janzert: Ah, yep that should work fine then.

PR #328, opened 2017-11-25
Make error detection in Haskell/Stack compilation more robust

### LINKED ISSUE #265: Haskell : stack support

lidavidm:
The logic behind the original `ErrorFilterCompiler` honestly makes little sense. This new compiler searches stderr for a particular regex; if it matches, it'll provide all of stdout and stderr, hopefully making debugging easier.

Fixes #265.

### DIFF
--- a/apiserver/worker/compiler.py
+++ b/apiserver/worker/compiler.py
@@ -269,6 +269,43 @@ class ErrorFilterCompiler(ExternalCompiler):
         return cmd_errors
 
 
+class ErrorSearchCompiler(ExternalCompiler):
+    """
+    A compiler that returns an error if stderr matches a certain regex.
+    """
+    def __init__(self, args, separate=False, out_files=[], out_ext=None,
+                 # Default regex matches nothing
+                 error_re="a^"):
+        ExternalCompiler.__init__(self, args, separate, out_files, out_ext)
+        self.error_re = re.compile(error_re)
+
+    def __str__(self):
+        return "ErrorSearchCompiler: %s" % (' '.join(self.args),)
+
+    def cmd_error_filter(self, cmd_out, cmd_errors):
+        result = [
+            "Error compiling with command {}".format(''.join(self.args)),
+            "stdout was:",
+        ]
+        result.extend(line for line in (cmd_out or []) if line is not None)
+        result.append("stderr was:")
+
+        found_error = False
+        for line in cmd_errors:
+            if line is None:
+                continue
+
+            if self.error_re.search(line):
+                found_error = True
+
+            result.append(line)
+
+        if found_error:
+            return result
+
+        return []
+
+
 class TargetCompiler(Compiler):
     def __init__(self, args, replacements, outflag="-o"):
         self.args = args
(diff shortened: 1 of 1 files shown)


Janzert:
So if I'm reading things correctly `ErrorFilterCompiler` filters out lines that are known to be ok and if there is anything left over it fails. This adds `ErrorSearchCompiler` which looks for known output and fails if it is found. Both seem to pretty fragile and error prone with likely false failures from `ErrorFilterCompiler` and false success from `ErrorSearchCompiler`.

Don't the compilers (the actual language compilers/build systems) have sensible exit codes that could be checked, e.g. anything non-zero is a failure?

But if not I suppose something like `ErrorSearchCompiler` is going to be the best we can do.

lidavidm:
Alright, changed to use the return code.

apiserver/worker/compiler.py
```diff
@@ -269,6 +269,31 @@ def cmd_error_filter(self, cmd_out, cmd_errors):
...
+            return []
+
+        result = [
+            "Error compiling with command {}".format(''.join(self.args)),
+            "Process exited with return code {}".format(returncode),
+            "stdout was:",
+        ]
+        result.extend(line for line in (cmd_out or []) if line is not None)
+        result.append("stderr was:")
+        result.extend(line for line in (cmd_errors or []) if line is not None)
```
Janzert: I know there is a comment elsewhere that it can happen and is protected against many places so it's fine to do so here as well. But do you know how and what causes a line to end up `None` instead of a string?

From what I can see `cmd_out` and `cmd_error` are always the product of a `string.split("\n")` with some items (lines) possibly removed. That shouldn't be adding `None` items.

lidavidm: I'm not sure, either. It comes from 5c995e6f2 but there's no indication of why there, either.

apiserver/worker/compiler.py
```diff
@@ -269,6 +269,31 @@ def cmd_error_filter(self, cmd_out, cmd_errors):
...
+
+    def __str__(self):
+        return "ReturncodeCompiler: %s" % (' '.join(self.args),)
+
+    def cmd_error_filter(self, cmd_out, cmd_errors, returncode):
+        if returncode == 0:
+            return []
+
+        result = [
+            "Error compiling with command {}".format(''.join(self.args)),
```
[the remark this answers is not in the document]
lidavidm: Whoops, you're right.

PR #329, opened 2017-11-27
dlang: statically disable logger during compilation

AurelC2G:
Adding the flag -version=StdLoggerDisableLogging to the compiler will statically disable all logging for bots written in D (see https://dlang.org/phobos/std_experimental_logger.html), improving the speed of the final build. This allows uploading a bot without scrubbing away logging used during development.


harikmenon:
@AurelC2G do you want to submit a starter kit for Dlang?

PR #333, opened 2017-11-27
Calculate total ships in game.

Janzert:
I forgot about the initial ships for each player and was using ships produced as the total number of ships. This corrects it by adding the initial ships to the total in a game.

### DIFF
--- a/website/javascript/templates/GameFeed.vue
+++ b/website/javascript/templates/GameFeed.vue
@@ -41,7 +41,7 @@
             </td>
             <td class="text-center hidden-xs">
               <div class="td-wrapper"><div class="text-center">
-                {{ game.ships_destroyed }} of {{ game.ships_produced }}
+                {{ game.ships_destroyed }} of {{ game.total_ships }}
                 <span v-if="game.planets_destroyed">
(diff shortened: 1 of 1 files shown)


harikmenon:
@Janzert I'm merging this in but the new deplopyment is gated on finishing up the basic work for challenges, dont want to cherry pick changes. The intial version of game feed is live now.

website/javascript/templates/GameFeed.vue
```diff
@@ -114,8 +114,10 @@ export default {
                 const players = Object.values(game.players).sort((r1, r2) => {
                   return r1.rank - r2.rank
                 })
-
                 game.playerSorted = players
+
+                // include the initial 3 ships of each player
+                game.total_ships = game.ships_produced + (players.length * 3);
```
Kerollmops: Why is it hard coded ? 3 is not in a global variable or something ?

Janzert: Mostly since so far as I know the game constants aren't available at this point, it's unlikely to be needed anywhere else besides this line and is very unlikely to change. But if someone wants to pull it out into a constant feel free.

PR #336, opened 2017-11-27
Change game id type so it won't overflow.

Janzert:
*** This PR should be left unmerged until such time it is ready to be run on the production database. ***

This changes the game id type in the database from mediumint to int so that the field doesn't overflow after 16 million games.

Because mysql cannot alter multiple tables in one transaction, before running the alembic upgrade game updates to the database should be stopped. This is probably easiest done by either by stopping the coordinator or stopping all workers. Specifically because the on delete cascade constraints will not be enforced while the update is made any games deleted during the update will leave orphan entries in the dependent tables. This will cause the re-addition of the constraints to fail after the update completes.

If other alembic revisions are added after the creation of this PR but before it is merged the revision here will need to have it's down_revision changed to the current head.

### DIFF
--- a/apiserver/alembic/versions/201711272050_5c3b1509e65d_change_game_id_type.py
+++ b/apiserver/alembic/versions/201711272050_5c3b1509e65d_change_game_id_type.py
@@ -1,0 +1,87 @@
+"""Change game id type.
+
+Revision ID: 5c3b1509e65d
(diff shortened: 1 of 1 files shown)


Janzert:
Sorry thought I was looking at my div end tag PR #337, but I see @harikmenon just merged it.

PR #341, opened 2017-12-03
Increase pipe buffer size on Linux

### LINKED ISSUE #340: Engine terminates bots if serialized game state is over 65536 bytes
2^16 being the size of a pipe's buffer, of course.

lidavidm:
The default pipe buffer size is 65536 bytes on Linux. If there are
a lot of ships, the serialized map state might be larger than that,
so we will encounter an error when trying to write it, and thus we
will eject all bots. On Linux, we can resize the buffer to avoid
this problem. This sets the size of the buffer to the maximum
size allowable (which is 1 MiB on my system) on Linux.

The alternative is to write the map in pieces, waiting for the
bot to read and clear the buffer before writing more. However,
then we either have to count the time to write the map as part of
the bot's turn time, or we have to introduce a separate timeout.
Otherwise, a bot could intentionally let its pipe fill up to buy
extra processing time while the engine waits for it to read the
map. And if we add a separate timeout, I fear bots will have to
implement this pathological behavior, because it's a way to
maximize the amount of time you get per turn.

This does not address OSX or Windows.

Fixes #340.

### DIFF
--- a/environment/networking/Networking.cpp
+++ b/environment/networking/Networking.cpp
@@ -218,6 +218,7 @@ void Networking::send_string(hlt::PlayerId player_tag,
     }
 #else
     UniConnection connection = connections[player_tag];
+
     ssize_t charsWritten =
         write(connection.write, sendString.c_str(), sendString.length());
     if (charsWritten == -1) {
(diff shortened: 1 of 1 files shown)


Janzert:
If I didn't mess up the math anywhere along the way a 1MB buffer should be enough for ~15.5k ships on the map. Also under ideal conditions the most ships that can be produced in a 300 turn game is a little under 3200. So this should be plenty large enough to ensure that the buffer never overflows.

PR #342, opened 2017-12-03
max_frame_response_time statistic

fohristiwhirl:
This attempts to add a max_frame_response_time statistic to the replay JSON.

However I don't speak C++ nor do I have a working C++ compiler here, so I don't even know if it compiles...


fohristiwhirl:
My rationale for this is that it would be nice to be able to find out how close to timeouts one is skating, on the official server. (Locally one can get the bot to log such info.)

lidavidm:
Travis failure seems fine, looks like one of the CI machines lost Internet during `apt-get update`.

PR #354, opened 2017-12-14
Add winter's and New Year's planets

lidavidm:
download 17
download 16


lidavidm:
@julskast how does this look?

julskast:
Looks nice @lidavidm.  @j-clap which one do you like better? I say the first one is clearer - definitely winter!

lidavidm:
Right now it (should) automatically switch over once the new year starts,
as another option.

j-clap:
I like the [first one](https://user-images.githubusercontent.com/327919/33968851-2b0aa5dc-e038-11e7-840e-b2bac31f9bc6.png) the most. They're loverly.

julskast:
@lidavidm switch over from what to what and on what day?

I'd be ok with using the winter one from Dec 23-Dec 30 and the other second one from Dec 31-Jan 1, but only if that's easy to do. Otherwise should just be the winter one (the first one) from Dec 23-Jan 1.

Thanks!

lidavidm:
Right now it changes from the winter image to the New Year's image on midnight Jan 1, but I can adjust the dates.

julskast:
Awesome, so yeah @lidavidm let's adjust the dates as I specified if you don't mind. Other than that, good to go!

lidavidm:
No problem, it's been updated!

PR #355, opened 2017-12-14
Implement various things for finals

Janzert:
This adds a few things for the finals to start with a full leaderboard reset and run with a progressive rank cutoff.

Main changes implemented are:
  * Adds a script to do a leaderboard reset by storing a new bot history entry with the current rating information and then resetting the rank, rating and games played information. This allows the final open competition rank and rating to be checked as the last retired entry in the bot history.
  * Set tau to 0 for finals as the composition of entries will be fixed.
  * Implements a method to set a schedule for the rank cutoff to automatically change as games are played in the finals.

I put example schedules for around 4000 and 5000 players in `config.py`. At current rates it looks like it will probably be around 3200 to 4200 players in the finals. As they get closer and the actual number of players can be predicted more accurately the schedule can be set accordingly. It's also pretty flexible on how large a proportion to give the top bots and where to set that final cutoff. The examples set the last cutoff for silver tier and above to get a few thousand games more than the next step down.

The example schedules also use a somewhat conservative 3000 games per hour where the actual rate has averaged around 3200 and they will probably run somewhat more than 6 full days.

Still on the to do list is a finals status page showing the current cutoff, number of games played in finals, cutoff schedule, and anything else that might be of interest.

### DIFF
--- a/apiserver/apiserver/config.py
+++ b/apiserver/apiserver/config.py
@@ -4,6 +4,34 @@ from passlib.context import CryptContext
 COMPETITION_OPEN = True
 # Original PHP: "compState", "finalsPairing"
 COMPETITION_FINALS_PAIRING = False
+# Final open game id, set this to the id of last game played in the open stage
+LAST_OPEN_GAME = None
+# Rank cutoff schedule during finals
+# In each entry the first value is number of games to start the cutoff and
+# second value is the rank cutoff to use.
+# Following schedule is based on ~4000 bots for 6 days of ~3000 games per hour
+FINALS_CUTOFF_SCHEDULE = [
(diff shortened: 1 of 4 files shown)


Janzert:
I'll need to dig out the checklist from last year to make sure there wasn't more, but at the very least the above list needs to add a couple items to the beginning:

* Set the `COMPETITION_OPEN = False` in `config.py` and then restart the coordinator server(s).
* Wait for all compilation jobs to finish
* Then stop all workers and coordinator
* set `LAST_OPEN_GAME` and so on

Basically closing the open competition is a two stage process of stopping submits and then allowing any compilation jobs to finish.

PR #358, opened 2017-12-14
Update Clojure starter kit to be more idiomatic.

grzm:
Pass the game map as an argument to bot functions rather
than binding and referencing global values.


grzm:
Please note this doesn't change the logic of the starter bot at all. It primarily removes the use of globals variables.

PR #362, opened 2017-12-17
Get rid of drag entirely (fixes #114)

### LINKED ISSUE #114: Non-zero velocities causing docking to fail

lidavidm:
Just explicitly set ship velocities to 0 at the end of a turn.
@ruilov found a case where a thrust of 7 @ 271 degrees produced
a velocity with a magnitude slightly higher than 7, so even with
the correct deceleration, there would still be residual velocity.

### DIFF
--- a/environment/core/Halite.cpp
+++ b/environment/core/Halite.cpp
@@ -360,18 +360,12 @@ auto Halite::process_production() -> void {
 
 auto Halite::process_drag() -> void {
     // Update inertia/implement drag
-    const auto drag = hlt::GameConstants::get().DRAG;
     for (auto& player_ships : game_map.ships) {
         for (auto& pair : player_ships) {
             auto& ship = pair.second;
-
-            const auto magnitude = ship.velocity.magnitude();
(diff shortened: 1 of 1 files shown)


ruilov:
Could you allow docking whenever the speed is “close enough” to zero? (And at docking events also explicitly set the speed to zero)

lidavidm:
@j-clap We can set drag to be higher than max speed, instead of equal to max speed; that should catch any rounding errors.

j-clap:
@lidavidm I'd definitely prefer that :)

lidavidm:
Closing in favor of #371.

PR #372, opened 2017-12-21
Shuffle player order for games

lidavidm:
Makes sure the seed player isn't always player 0; hopefully sidesteps any bias in starting positions. Based on some discussion in Discord.


Janzert:
I'd rather not do this. It's handy to know who the seed player is, and in fact the game feed page currently highlights them with a yellow bar over the icon.

lidavidm:
Ah, I see. There was some concern about the tiebreaker still, though it's unclear what's going on (dhallstr always has player 0 win in different seeds, but nobody else can reproduce it).

lidavidm:
Alright, let's first see what's going on with the tiebreaker before we consider this.

PR #386, opened 2017-12-27
Don't count challenge games.

Janzert:
Challenge games shouldn't count in the number of games played since they don't follow the standard matchmaking and don't update the rating.

### DIFF
--- a/apiserver/apiserver/coordinator/coordinator.py
+++ b/apiserver/apiserver/coordinator/coordinator.py
@@ -240,13 +240,14 @@ def store_game_results(conn, game_output, stats, replay_key, bucket_class,
             sigma=user["sigma"]
         ))
 
-        # Increment number of games played
-        conn.execute(model.bots.update().where(
-            (model.bots.c.user_id == user["user_id"]) &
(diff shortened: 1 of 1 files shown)


apiserver/apiserver/coordinator/coordinator.py
```diff
@@ -240,13 +240,14 @@ def store_game_results(conn, game_output, stats, replay_key, bucket_class,
...
         ))
 
-        # Increment number of games played
-        conn.execute(model.bots.update().where(
-            (model.bots.c.user_id == user["user_id"]) &
-            (model.bots.c.id == user["bot_id"])
-        ).values(
-            games_played=model.bots.c.games_played + 1,
-        ))
+        if not challenge:
```
lidavidm: I think this would count games with challenge ID 0, but I already used that, so oh well.

Janzert: Yeah, it would and so would the same check at line 144 to not update the ratings. But like you said since challenge 0 is already past it shouldn't matter now.

PR #392, opened 2017-12-28
Return compilation status in leaderboard api.

Janzert:
This will allow eventually either filtering out or marking bots on the leaderboard that aren't actively playing games.

### DIFF
--- a/apiserver/apiserver/web/leaderboard.py
+++ b/apiserver/apiserver/web/leaderboard.py
@@ -122,6 +122,7 @@ def leaderboard():
                 "organization_id": row["organization_id"],
                 "organization": row["organization_name"],
                 "version_number": int(row["num_submissions"]),
+                "compile_status": row["compile_status"],
                 "num_games": int(row["num_games"]),
                 "score": float(row["score"]),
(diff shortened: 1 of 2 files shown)


apiserver/apiserver/web/leaderboard.py
```diff
@@ -122,6 +122,7 @@ def leaderboard():
                 "organization_id": row["organization_id"],
                 "organization": row["organization_name"],
                 "version_number": int(row["num_submissions"]),
+                "compile_status": row["compile_status"],
```
lidavidm: If you add `"compile_status": model.ranked_bots_users.c.compile_status` to the `get_sort_filter` call above, then the API should also let people filter/sort on the compile status field.

Janzert: I was going to add that and still plan too. But while I didn't actually test it I don't think `get_sort filter` currently handles enumeration types? So it'll take a little work to add that as well.

lidavidm: Ah whoops, you're right, was thinking it was just a string column for some reason.

PR #396, opened 2017-12-29
Add API endpoint to list all challenges

lidavidm:


apiserver/apiserver/web/challenge.py
```diff
@@ -0,0 +1,156 @@
...
+            model.challenges.c.winner,
+        ]).select_from(model.challenges.join(
+            model.challenge_participants,
+            (model.challenges.c.id == model.challenge_participants.c.challenge_id) &
+            participant_clause
+        )).where(
+            where_clause
+        ).order_by(
+            *order_clause
+        ).offset(offset).limit(limit).reduce_columns()
```
Janzert: Take the following with the caveat that I may be seeing this incorrectly here since I'm manually adding the challenge data to my database to test with.

In the case of no extra participant_clause it seems like this query will return every challenge multiple times, i.e. once for each participant because of the join. But rather than fix this query to only return one row per challenge, it might be better to expand it to include the participant columns and then format the json result from there. This would do away with the next section that is querying the database for every participant for each challenge. That way it would be 1 database query instead of 1+total_number_of_challenge_participants queries. I think this also requires moving the participant_clause from being a join condition to a where condition.

lidavidm: You're right about the duplicates. Pushed a fix that should only use one query.

PR #405, opened 2017-12-31
Make game deletion script executable

lidavidm:
Cron job doesn't seem to actually be running, this should let it run.


lidavidm:
Also fixes it deleting games for users with rank >= bottom Bronze tier rank instead of > bottom Silver tier rank

PR #406, opened 2017-12-31
Add Cython starter kit

rlee287:
This is a Cython version of the Python starter kit. The bottleneck functions and methods were Cythonized by adding static types.


harikmenon:
Not a Cython expert. Overall looks good. Thanks for the contribution. Will deploy this to the site tomorrow

PR #408, opened 2018-01-03
Select challenge winner only from actual participants; only issues 4-player games in 4-player challenge

lidavidm:
Thanks to @fohristiwhirl for finding this. 🤦‍♂️


lidavidm:
Will also come up with a query to fix the existing challenges in a bit.

lidavidm:
Also changes it so that 4-player challenges only have 4-player matches.

Issue #402, opened 2017-12-29
Paused challenges


julskast:
Sometimes players on the leaderboard can't play in a challenge. E.g. opponents with currently disabled bots (due to compilation) won't have a bot to play.

- [ ] Let's add a state to the analysis widget for 'paused' if at least one player in the challenge does not have a bot. 

E.g.
![image](https://user-images.githubusercontent.com/764613/34448195-310a8906-ecb9-11e7-9e38-9e1e2170d571.png)

toannguyen132:
@harikmenon how can I get the pause status in the response of server? All I can see in the `finished ` key of the response is bool value (true or false).

julskast:
@lidavidm any suggestions here?

lidavidm:
We don't explicitly return it in the API, but you can check the compilation status of the participants' bots (or whether they even have uploaded a bot yet).

```
$ curl https://api.halite.io/v1/api/user/1000/bot/0
{
  "bot_id": 0, 
  "compilation_status": "Successful", 
  "games_played": 562, 
  "language": "Python", 
  "rank": 942, 
  "score": 26.6559, 
  "version_number": 14
}
```

julskast:
Thanks @lidavidm ! @toannguyen132 let us know if this makes sense. cc @benwong0

toannguyen132:
@lidavidm so I need to fetch all the participants' information in each challenge in the list to get the compilation status in order to display?

lidavidm:
@toannguyen132 Yes, all the bots of all the participants, if the challenge is not completed. If that's too many requests, then let's just add it to the API.

toannguyen132:
@lidavidm I have updated the paused status. But I do not know any user those don't have any bot or failed compilation to test. I tried to fail my bot by uploading the wrong bot but it still shows my bot status 'successful'. I think it uses my previous version.
Do you have any example user like that?

lidavidm:
@toannguyen132 I just updated my bot to fail (user 1000), does that help?

Issue #394, opened 2017-12-28
Turkish characters


julskast:
Kırıkkale is shown as K?r?kkale

- [ ] Can we make sure we support Turkish characters as an org name on the leaderboard, user profile, etc.?

toannguyen132:
@julskast the content font is “helvetica neue” and users may not have it. I currently use web-font (client will download font) to display it correctly. Should I just declare ‘helvetica neue’ and if the users don’t have it, they will see default font instead? I think this is the best way to fix this issue.

julskast:
@toannguyen132 sounds good to me! thanks!

toannguyen132:
@julskast can you please provide me a screenshot of it. Where did you get that issue?

julskast:
<img width="1191" alt="screen shot 2018-01-03 at 4 43 09 pm" src="https://user-images.githubusercontent.com/764613/34541164-3a2cc536-f0a5-11e7-8d29-4ae952b007b4.png">

If you filter the leaderboard by "Turkey" the first name has ? in it where the special characters should be.

https://halite.io/programming-competition-leaderboard?country=TUR

toannguyen132:
@julskast Those characters are originally returned from the server. I think this is not the font issue.

<img width="356" alt="returned characters" src="https://user-images.githubusercontent.com/7789414/34553807-583fe024-f0de-11e7-9078-573fe179e1d2.png">

julskast:
![image](https://user-images.githubusercontent.com/764613/34570120-e003cb42-f138-11e7-9ad4-291f3045a154.png)

Perhaps it's both front end and backend? Here's the user's profile https://halite.io/user/?user_id=6169

@lidavidm any thoughts here? would be nice to support these turkish characters if we can, but we obviously can't support all characters in every language...

lidavidm:
It's a configuration issue in the backend. I'll let Hari know for the next redeployment.

harikmenon:
Closing issue since competition is ended

PR #412, opened 2018-01-04
Update font, paused challenges #402 #394

toannguyen132:
#394, #402 
update pause challenges


website/javascript/templates/UserProfile.vue
```diff
@@ -814,18 +812,23 @@
...
+                    }
+                    // bots compilation is failed
+                    else if (this.participants[p.user_id] &&
+                      this.participants[p.user_id][0].compilation_status !== 'Successful' &&
+                      !c.finished){
+                        challenges[i].status = 'Paused'
+                        flag = true
+                      } else {
+                        challenges[i].status = c.finished ? 'Completed' : 'In Progress'
+                      }
```
Janzert: It seems like the check for challenged finished should be before the checks to see if the challenge is paused. In other words. once finished the challenge should always be 'Completed' no matter what the state of the participants bots is in at that point.

PR #426, opened 2018-01-09
Add date as argument user mode of the replay mode in the hlt client

iamtomcheng:
This is a quick fix by adding the date as argument in the respective methods and classes and adding an if statement to skip replays that belong to the wrong date in case the requested date is given as input. Please feel free to amend it to make it more efficient.

### DIFF
--- a/tools/hlt_client/hlt_client/client.py
+++ b/tools/hlt_client/hlt_client/client.py
@@ -184,6 +184,9 @@ def _parse_arguments():
                                     help='Number of replays to fetch')
     replay_user_parser.add_argument('-d', '--destination', dest='destination', action='store', type=str, required=True,
                                     help="In which folder to store all resulting replay files.")
+    replay_regex_parser.add_argument('-t', '--date', action='store', type=str, dest='date', required=True,
+                                     help="Fetch replay files matching the specified date. To fetch a day's files user"
+                                          "the YYYYMMDD format.")
     # .Modes.Replay.Modes.Date
     replay_regex_parser = replay_subparser.add_parser(REPLAY_MODE_DATE, help='Retrieve replays based on regex')
     replay_regex_parser.add_argument('-t', '--date', action='store', type=str, dest='date', required=True,
--- a/tools/hlt_client/hlt_client/client.py
+++ b/tools/hlt_client/hlt_client/client.py
@@ -184,7 +184,7 @@ def _parse_arguments():
                                     help='Number of replays to fetch')
     replay_user_parser.add_argument('-d', '--destination', dest='destination', action='store', type=str, required=True,
                                     help="In which folder to store all resulting replay files.")
-    replay_regex_parser.add_argument('-t', '--date', action='store', type=str, dest='date', required=True,
+    replay_user_parser.add_argument('-t', '--date', action='store', type=str, dest='date', required=True,
                                      help="Fetch replay files matching the specified date. To fetch a day's files user"
                                           "the YYYYMMDD format.")
     # .Modes.Replay.Modes.Date
--- a/tools/hlt_client/hlt_client/download_game.py
+++ b/tools/hlt_client/hlt_client/download_game.py
@@ -128,30 +128,41 @@ class UserGameDownloader(GameDownloader):
     _FETCH_THRESHOLD = 250
     _BUCKETS = []
 
-    def __init__(self, destination, user_id, limit):
+    def __init__(self, destination, user_id, limit, date):
         """
         Download games for a user
         :param destination: Where to download
         :param user_id: Which user's replays to fetch
         :param limit: How many replays to fetch (max)
+        :param date: Which date to download
         """
         self.destination = destination
-        self.objects = self._parse_user_metadata(self._fetch_metadata(user_id, limit))
+        self.objects = self._parse_user_metadata(self._fetch_metadata(user_id, limit, date))
 
-    def _fetch_metadata(self, user_id, limit):
+    def _fetch_metadata(self, user_id, limit, date):
         """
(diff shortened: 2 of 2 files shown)


tools/hlt_client/hlt_client/client.py
```diff
@@ -184,6 +184,9 @@ def _parse_arguments():
                                     help='Number of replays to fetch')
     replay_user_parser.add_argument('-d', '--destination', dest='destination', action='store', type=str, required=True,
                                     help="In which folder to store all resulting replay files.")
+    replay_user_parser.add_argument('-t', '--date', action='store', type=str, dest='date', required=True,
```
j-clap: This should not be a required argument.

tools/hlt_client/hlt_client/download_game.py
```diff
@@ -128,30 +128,41 @@ class UserGameDownloader(GameDownloader):
...
         while current <= limit:
-            current_limit = self._FETCH_THRESHOLD if ((limit - current) >= self._FETCH_THRESHOLD) else (limit - current)
-            result_set += requests.get(self._USER_BOT_URI.format(user_id, current_limit, current)).json()
-            current += self._FETCH_THRESHOLD
+            if date is None:
+                current_limit = self._FETCH_THRESHOLD if ((limit - current) >= self._FETCH_THRESHOLD) else (limit - current)
+                result_set += requests.get(self._USER_BOT_URI.format(user_id, current_limit, current)).json()
+                current += self._FETCH_THRESHOLD
+            else:
+                if requests.get(self._USER_BOT_URI.format(user_id, 1, current)).json()[0]["replay"].split("-")[1] != date:
```
j-clap: Magic number

tools/hlt_client/hlt_client/download_game.py
```diff
@@ -128,30 +128,41 @@ class UserGameDownloader(GameDownloader):
...
+                    current += 1
+                    limit   += 1
+                    continue
+                else:
+                    result_set += requests.get(self._USER_BOT_URI.format(user_id, 1, current)).json()
```
j-clap: Magic number

tools/hlt_client/hlt_client/client.py
iamtomcheng: Okay, I will change it to 'default=None' and remove 'required=True'.

tools/hlt_client/hlt_client/download_game.py
iamtomcheng: I was trying to check the date of the retrieved replay one at a time. Do you have any suggestion on how to do this faster? I can also retrieve current_limit replays at a time and filter out replays that belongs to the wrong date and append the filtered list to the list result_set.

Janzert:
You probably don't want to be filtering on the date client side, rather use the server side filtering. For example to get the games for iamtomcheng on new years day use a url something like this:

`https://api.halite.io/v1/api/user/4882/match?limit=250&filter=time_played,>=,2018-01-01T00:00&filter=time_played,<,2017-01-02T00:00`

You'll still generally need to make multiple requests with appropriate limit and offsets to get the complete day.

Also if you do want to do any date filtering on the client side it'd probably be best to use the `time_played` field returned with the metadata rather than parsing it out of the replay name.

iamtomcheng:
@Janzert So I tried your url and it seems like the second filter on time_played is not working, i.e. 250 replays starting from 2018-01-01T00:00 are returned, and also you put 2017 instead of 2018. I removed the second filter on time_played, add the offset back and add a client side filter using the time_played field to keep the logic simple.

Janzert:
Sorry, so besides the obvious date typo it seems like when multiple filters of the same type are applied they get combined as `first_filter OR second_filter`. So the above query ends up matching for every game.

Filters of different types (e.g. game_id and time_played) do get applied with AND, so I have no idea what is going on.

And yes no matter what any single query will only return a maximum of 250 games, that's what I meant about needing to use multiple requests with appropriate limit and offsets.

iamtomcheng:
Hi @j-clap, could you take a look at the modifications and see if there is still any other issue? Thanks.

tools/hlt_client/hlt_client/download_game.py
```diff
@@ -125,32 +126,40 @@ def __init__(self, destination, date, all_bots=False):
...
         print('Fetching Metadata')
         current = 0
         result_set = []
         while current <= limit:
             current_limit = self._FETCH_THRESHOLD if ((limit - current) >= self._FETCH_THRESHOLD) else (limit - current)
-            result_set += requests.get(self._USER_BOT_URI.format(user_id, current_limit, current)).json()
+            if date is None:
+                result_set += requests.get(self._USER_BOT_URI.format(user_id, current_limit, current)).json()
+            else:
+                requested_date = datetime.datetime.strptime(date,'%Y%m%d').strftime('%Y-%m-%dT00:00')
```
Janzert: Btw, if you're always going to filter at the level of a full day you can drop the `T00:00`.

iamtomcheng: Right, I just dropped the T00:00.

iamtomcheng:
Hi @harikmenon, could you take a look at the changes made and see if it is okay now. Also, as pointed out by @Janzert, there is a bug (?) in the Halite API that when multiple filters of the same type are applied they get combined as `first_filter OR second_filter` instead of `first_filter AND second_filter`. Maybe you would like to fix that as well if it was not intentional.

harikmenon:
@j-clap can you review?

Janzert:
To be clear the OR'ing of multiple uses of a filter in the api is definitely intentional. I'm not sure when the capability is used, but it is special cased in the code to make it OR instead of AND. Certainly both are useful in different situations.

PR #430, opened 2018-01-11
Post-finals challenge mode

lidavidm:
Note for deployment: this adds a new configuration setting.

### DIFF
--- a/apiserver/apiserver/coordinator/compilation.py
+++ b/apiserver/apiserver/coordinator/compilation.py
@@ -145,6 +145,12 @@ def update_compilation_status():
                     )
                 )
 
+            # In challenge mode, do not reset sigma.
+            if config.COMPETITION_CHALLENGE_MODE:
+                new_sigma = model.bots.c.sigma
+            else:
+                new_sigma = 8.333
+
             # We do not reset the bot rank for new submissions now.
             conn.execute(
                 model.bots.update().where(
@@ -154,7 +160,7 @@ def update_compilation_status():
                     language=language,
                     version_number=model.bots.c.version_number + 1,
                     games_played=0,
-                    sigma=8.333,
+                    sigma=new_sigma,
                 )
             )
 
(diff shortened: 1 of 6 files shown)


apiserver/apiserver/coordinator/compilation.py
```diff
@@ -145,6 +145,12 @@ def update_compilation_status():
                     )
                 )
 
+            # In challenge mode, do not reset sigma.
+            if config.COMPETITION_CHALLENGE_MODE:
+                new_sigma = model.bots.c.sigma
```
j-clap: Are you sure about this? Some will complain about the decrease

j-clap: Disregard that statement, I misread the reason. This is fine

apiserver/apiserver/coordinator/compilation.py
```diff
@@ -154,7 +160,7 @@ def update_compilation_status():
                     language=language,
                     version_number=model.bots.c.version_number + 1,
                     games_played=0,
-                    sigma=8.333,
+                    sigma=new_sigma,
```
Janzert: Still working through this, but I'm thinking games_played shouldn't be reset either. Not positive on that, but I think it makes sense since we have challenge games avoid incrementing that count now.

lidavidm: Hmm, we are incrementing the version number though, and it is a different bot. Maybe it'd make more sense to have challenge games increment the count in this mode.

Janzert: I guess it depends somewhat on, is the number of games you played in the finals going to be a point of pride or 'badge' of any sort for people going forward? I notice that it is one of the three things we display on current profiles as a stat from last year. But then of course so is number of submissions, i.e. version number. *shrug*

harikmenon:
I'm going to close this. We are not going tp allow challenges after finals :"(

PR #433, opened 2018-01-15
Catch IntegrityError for duplicate bot history entries

lidavidm:
https://forums.halite.io/t/am-i-missing-something-obvious-or-how-do-i-increment-bot-version/1295/6

### DIFF
--- a/apiserver/apiserver/coordinator/compilation.py
+++ b/apiserver/apiserver/coordinator/compilation.py
@@ -131,19 +131,27 @@ def update_compilation_status():
                     ]).select_from(model.users)
                 ).first()[0]
 
-                conn.execute(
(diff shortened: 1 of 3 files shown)


Janzert:
This looks like it will correct the situation where the bot_version number gets out of sync.

In order to prevent it from getting out of sync in the first place you may want to have everything execute in one transaction so that if the version number update fails (or anything else) the history insert gets rolled back as well. Fortunately that is very easy to accomplish just change line 68 to call `model.engine.begin()` instead of `connect()`.

lidavidm:
Good idea, thanks.

I've also exposed some more fields for filtering in the challenges API endpoint.

PR #435, opened 2018-01-15
Increase limit and ensure most recent challenges are shown.

Janzert:
This increases the limit for number challenges displayed from 50 to 200. That should be enough to show the maximum number of challenges between two players in a month. Also specify the order of challenges received to ensure the most recent challenges are shown.


Janzert:
Already done in #434

PR #439, opened 2018-01-15
Require that all bots in a challenge are able to play a game

### LINKED ISSUE #438: 4 player challenges can still play 2 player games
See challenge 1174 for an example.
http://api.halite.io/v1/api/challenge/1174
http://api.halite.io/v1/api/challenge/1174/match

Most of the games are played only between the issuer and tscommander-ml.

In this case I believe it is because tsadmiral and tscaptain have is_gpu_enabled turned on. So they only play if the challenge happens to be selected to play on a gpu enabled server.

I believe it's also the case that if one of the challenged bots had a failed upload or was otherwise disabled the remaining two challenged bots would continue playing 2 player games against the issuer.

lidavidm:
Fixes #438.

### DIFF
--- a/apiserver/apiserver/coordinator/matchmaking.py
+++ b/apiserver/apiserver/coordinator/matchmaking.py
@@ -187,10 +187,6 @@ def find_challenge(conn, has_gpu=False):
     if not challenge:
         return None
 
-    player_filter = sqlalchemy.sql.expression.true()
-    if not has_gpu:
-        player_filter = (model.ranked_bots_users.c.is_gpu_enabled == False)
-
     total_players = conn.execute(model.total_ranked_users).first()[0]
     thresholds = util.tier_thresholds(total_players)
     ranked_users = model.ranked_users_query()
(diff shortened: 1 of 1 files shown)


Janzert:
I think this still doesn't handle the situation where some of the users have is_gpu_enabled set and the worker is not gpu enabled?

lidavidm:
Missed that, thanks.

lidavidm:
Hmm, though this change will mean that challenge games for GPU-enabled bots will become very infrequent.

Janzert:
I think in order to make it work and have a reliable rate of challenge games it needs to change from the current method of choosing a challenge then checking if the players are eligible to play.

Instead it should choose only a challenge that has all eligible players (successful compile and allowable gpu status). This will automatically tend to pick up challenges with gpu_enabled players whenever it is a gpu worker.

lidavidm:
Alright, I modified the initial query to select a challenge so that non-GPU workers ignore challenges involving GPU bots. Will have to modify it some more to also filter out bots that don't compile.

Janzert:
If it's difficult to include the bot filter in the initial query I think just taking what you have now is an improvement already. Since it looks like it will only be in place for a few more days at this point, it may not be worth the effort to get everything included.

lidavidm:
Good point, hopefully we can at least get this in then.

PR #444, opened 2018-01-18
Update finals status page design.

Janzert:
Updated page design from @julskast 

@julskast I did make a few more tweeks from the last one I showed you, so you'll probably want to look it over. I think the largest was adding the current number of games played in the finals to the matchmaking status.

During open competition:
open-competition

After submissions close, before finals start:
closed-submissions

Shortly after the games have started:
start

Middle stage:
mid

End stage:
end-stage

### DIFF
--- a/website/javascript/templates/FinalsStatus.vue
+++ b/website/javascript/templates/FinalsStatus.vue
@@ -1,48 +1,107 @@
 <template>
   <div class="finals-status">
-    <h2 v-if="submissions_open">Submissions are still open, keep those improvements coming.</h2>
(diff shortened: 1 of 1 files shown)


julskast:
@Janzert looks great! My only tweak would be to the hide the line at the bottom until the first group has been highlighted. This line: "* Note that the highlighted group..."

Other than that I think it's great!

julskast:
Awesome, thanks for that edit! @harikmenon gtg on my end, I'll let you guys take it from here

PR #456, opened 2018-01-21
julia starter kit: fixed logging for new Memento version, corrected closest_point_to calculation

ingenieurtoni:


ingenieurtoni:
closest_point_to(target position) was not between both entities but "behind" the target entity.

PR #457, opened 2018-01-22
Don't allow GPU bots to be seeded on a non-GPU worker

lidavidm:


apiserver/apiserver/coordinator/matchmaking.py
```diff
@@ -63,6 +63,12 @@ def serve_game_task(conn, has_gpu=False):
         else:
             seed_filter = (rank_limit
                           & (model.ranked_bots_users.c.is_gpu_enabled == False))
+    elif not has_gpu:
+        # Even if we don't require that GPU workers only receive games
+        # seeded with GPU bots, we still don't want GPU bots to get
+        # seeded on a non-GPU worker.
+        seed_filter = (rank_limit
+                       & (model.ranked_bots_users.c.is_gpu_enabled == False))
```
Janzert: The overall checks collapse to:
```
    if config.ENFORCE_GPU_SEEDING and has_gpu:
        seed_filter = (limit the seeds to gpu enabled only)
    elif not has_gpu:
        seed_filter = (limit the seeds to non-gpu only)
```
Which is indeed what we want and this formulation should work fine.

lidavidm: Yeah I considered collapsing the branches but it would be more explicit what was happening this way.

PR #470, opened 2018-08-05
Add Common Lisp starter kit.

marcoheisig:
A friend of mine asked me to write a Common Lisp starter kit.  Well, here it is.

Apart from the different naming conventions of Common Lisp, the starter kit is pretty much identical to the others.  It would be great if you could merge this pull request and add an entry to the "Downloads and Starter Kits" section.


snaar:
I'm not very familiar with lisp, so I might be missing something obvious - I can't quite see the place where the bot can use the 60 seconds to analize the inital map before game starts?

marcoheisig:
Oh, good point!  I fixed that, players can now decide when to send the bot name.

snaar:
Merged - thanks! Note that we are not going to touch Halite2 website at least until we are "done" with Halite3. It's possible we'll never touch Halite2 website or we might do something - no concrete plans either way.

